PR #5, opened 2015-01-19
Restful model collection count

kamilbielawski:
In one of my projects I needed to display total number of threads in namespace. This pull request adds method `count` to `RestfulModelCollection` which allows to easily achieve this.


README.md
```diff
@@ -140,21 +140,27 @@ thread = namespace.threads.find('ac123acd123ef123')
 # (paginating 50 at a time until no more are returned.)
 namespace.threads.where(:tag => 'inbox').each do |thread|
   puts thread.subject
-end    
+end
```
khamidou: Thanks for the cleanups!

lib/restful_model_collection.rb
```diff
@@ -27,6 +27,11 @@ def each
       end
     end
 
+    def count
+      resp = RestClient.get(url, params: @filters.merge(view: 'count'))
+      JSON.parse(resp)['count']
```
khamidou: Could you refactor this code to be a little more like the rest of the SDK?
Notably, wrapping the response handling in a block and passing it to `Inbox.interpret_response` to handle errors.

See https://github.com/inboxapp/inbox-ruby/blob/master/lib/message.rb#L35 for an example

khamidou:
Hi Kamil,

Thanks for the pull-requests. It's almost there, you just need to address the two comments I made.
Karim

ps: Could you sign our Contributor Licence Agreement (https://www.inboxapp.com/cla.html)?

kamilbielawski:
Thanks for the feedback. I made the changes you requested and I also signed CLA.

PR #20, opened 2015-04-09
Add reply_to_message_id parameter to Draft

ajhodges:
To be used as documented in: https://www.nylas.com/docs/api#sending


emfree:
Thanks! Will release an updated version to Rubygems soon.

Issue #21, opened 2015-04-10
Delta streaming support


ajhodges:
It would be nice to have support for streaming delta updates via the Nylas API. I know this is not a simple feature to add since RestClient does not have support for long polling.

I have taken a stab at implementing this in my fork, and I will be creating a pull request with the changes.

khamidou:
Hi Adam,

Thanks for the pull request! I'll review it and get back to you quickly.

Karim

PR #22, opened 2015-04-10
Add support for delta streaming API to Namespace

ajhodges:
As proposed in #21 :

My implementation adds two new dependencies - yajl-ruby for parsing JSON streams and em-http-request for streaming via HTTP.

Requesting feedback on the following:
I have not handled any HTTP errors in delta_stream because I am not entirely sure which errors the /delta/streaming endpoint can throw.

I have not implemented any sort of recovery if the connection fails. Not sure if this should be a part of the delta_stream function or not.

I am welcome to other feedback and proposals for alternate methods. I have also tried implementing this with yajl-ruby's HttpStream class, but it does not support HTTPS.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,8 @@
 source 'https://rubygems.org'
 
 gem 'rest-client', '~> 1.7'
+gem 'yajl-ruby'
+gem 'em-http-request'
 
 # Add dependencies to develop your gem here.
 # Include everything needed to run rake, tests, features, etc.
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ You don't need to use this repo unless you're planning to modify the gem. If you
 
 - Ruby 1.8.7 or above. (Ruby 1.8.6 may work if you load ActiveSupport.)
 
-- rest-client, json
+- rest-client, json, yajl-ruby, em-http-request
 
 
 ## Example Rails App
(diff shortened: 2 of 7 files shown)


khamidou:
Hi Adam,

Thanks for the pull request! Sorry for being late, the last month was hectic. I've got a couple things I'd like you to address before I can greenlight it.

1. Regarding error handling, I think the right thing to do here would be to have `#delta_stream` catch all the errors raised by the connection and then re-raise them. This way we can give the calling function important information like the current cursor.

2. Also, I don't really understand why you chose to use `yajl` instead of the standard ruby JSON module.

Karim

ajhodges:
Sure thing!

1. I just added some basic error handling a couple days ago and updated my commit. https://github.com/nylas/inbox-ruby/pull/22/files#diff-84ccdfa4e08110177e35ef9ce93da020R193 I'm not sure if that's what you had in mind.

2. The reasoning behind this is because the standard ruby JSON module does not support parsing JSON from streams. Since all we are given from the streaming endpoint is a stream of JSON, we need to be able to parse an indefinite amount of JSON objects as they come in. The chunk that is yielded by the stream method of `EventMachine::HTTPRequest` is not necessarily going to be a valid JSON object, so we have to pass them to `yajl`, which then yields objects as they are parsed.
That is my understanding of the issue, at least. I could be wrong, this is the first time I've parsed a streaming JSON endpoint.

brandonhilkert:
Any update on this? We just found it and are interested in the steaming elements here.

khamidou:
Done! Thanks for the pull request, @ajhodges!

brandonhilkert:
:+1:  Nice @ajhodges, awesome work!

PR #25, opened 2015-07-03
Moved sample rails app to rails subfolder

KartikTalwar:
This pull request moves the sample rails app to a `rails` sub-folder inside the `example` folder in order to make space for more examples.


example/rails/config/environments/development.rb
```diff
@@ -2,6 +2,7 @@
   # Settings specified here will take precedence over those in config/application.rb.
   config.nylas_app_id = 'YOUR_APP_ID'
   config.nylas_app_secret = 'YOUR_APP_SECRET'
+  config.nylas_api_server = nil
```
khamidou: Why are you defining this? IIRC this setting has a default value in the SDK (if not it's definitely a bug). If you're doing this to show people how it's possible to use their own API server, you should probably add a comment.

KartikTalwar: Hey @khamidou without specifying that variable I get this: http://i.imgur.com/5phVZUa.png . Does this not happen for you?

Issue #34, opened 2015-09-07
Add parameter content_id to File


jockee:

jockee:
Merged.

PR #35, opened 2015-09-07
Add content_id parameter

jockee:
https://github.com/nylas/nylas-ruby/issues/34


jockee:
Signed the CLA.

khamidou:
Thanks for the fix!

Karim

PR #38, opened 2015-09-18
Refactor building the exclude_types query param

stevenharman:
Leverage Ruby to make this more functional, and shorter! We were also able to clean up some incidental duplication.

I also added a test for `exclude_types` to make sure I didn't break anything. :)

### DIFF
--- a/lib/delta_filters.rb
+++ b/lib/delta_filters.rb
@@ -7,7 +7,7 @@ module Inbox
       new.build_exclude_types(object_types)
     end
 
-    def initialize(filter_map: API::OBJECTS_TABLE)
+    def initialize(filter_map=API::OBJECTS_TABLE)
       @filter_map = filter_map
     end
 
--- a/lib/delta_filters.rb
+++ b/lib/delta_filters.rb
@@ -1,0 +1,23 @@
+module Inbox
+  class DeltaFilters
+
+    attr_reader :filter_map
+
+    def self.build_exclude_types(object_types)
+      new.build_exclude_types(object_types)
+    end
+
+    def initialize(filter_map: API::OBJECTS_TABLE)
+      @filter_map = filter_map
+    end
+
+    def build_exclude_types(object_types)
+      filters = Array(object_types).map { |type| filter_map.key(type) }.compact.join(',')
+
+      return filters if filters.empty?
+
+      "&exclude_types=#{filters}"
+    end
+
+  end
+end
--- a/lib/inbox.rb
+++ b/lib/inbox.rb
@@ -90,6 +90,20 @@ module Inbox
     attr_reader :app_id
     attr_reader :app_secret
 
+    OBJECTS_TABLE = {
+      'account' => Inbox::Account,
+      'calendar' => Inbox::Calendar,
+      'contact' => Inbox::Contact,
+      'draft' => Inbox::Draft,
+      'event' => Inbox::Event,
+      'file' => Inbox::File,
+      'folder' => Inbox::Folder,
+      'label' => Inbox::Label,
+      'message' => Inbox::Message,
+      'tag' => Inbox::Tag,
+      'thread' => Inbox::Thread,
+    }
+
     def initialize(app_id, app_secret, access_token = nil, api_server = 'https://api.nylas.com',
                    service_domain = 'api.nylas.com')
       raise "When overriding the Inbox API server address, you must include https://" unless api_server.include?('://')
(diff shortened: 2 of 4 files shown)


khamidou:
Hi Steven,

Thanks for the pull request! Your code looks good to me apart from a tiny nitpick: could you directly test `build_exclude_types`? For example, 

```ruby
describe "build_exclude_types" do
  it "returns an empty string when receiving an empty list" do
    qs = @inbox.send(:build_exclude_types, [])
    expect(qs).to eq('')
  end

  it "builds a correct query string" do
     qs = @inbox.send(:build_exclude_types, [Inbox::Message, Inbox::Thread])
    expect(qs).to eq(''&exclude_type=message,thread")
  end
end
```

Thanks.

Karim

stevenharman:
@khamidou, I consider it smell to be testing private methods. Luckily, we have a design tool at our disposal for such situations - extract a new object! That's what I've done. And it cleaned up the code a bit.

Thoughts?

stevenharman:
Oh, and it made it trivial to allow specifying a single object:

```ruby
nylas.deltas(cursor, Inbox::Message) #=> will exclude messages
```

lib/delta_filters.rb
```diff
@@ -0,0 +1,23 @@
+module Inbox
+  class DeltaFilters
```
khamidou: What about using a more generic name, like `DeltaQueryString` or `DeltaURLBuilder`?

lib/delta_filters.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    attr_reader :filter_map
+
+    def self.build_exclude_types(object_types)
+      new.build_exclude_types(object_types)
+    end
+
+    def initialize(filter_map: API::OBJECTS_TABLE)
```
khamidou: Please use a default parameter here (`filter_map=API:OBJECTS_TABLE`).

lib/delta_filters.rb
stevenharman: I thought about that, but really, we shouldn't be building URLs nor QueryStrings by hand. I was hoping to change the internals to leverage either `URI::Generic` (and friends) or `Addressable::URI` to do the actual URI building. Then this object's job would be purely about building the list of filters.

lib/delta_filters.rb
stevenharman: Any reason you prefer a default value vs. a keyword argument with a default value?

khamidou: Well, we still need to be compatible with Ruby 1.9.x. We still have customers using it.

khamidou:
Thanks for the update, Steven!

I left a couple comments inline. Here’s a bit more context. The delta stream API takes a number of parameters:
- exclude_types to exclude certain types.
- include_types to include only certain types.
- view=expanded to expand messages (i.e: expand messages have a couple additional headers like `Message-Id`. This is only if you need to thread messages together).

The Ruby SDK doesn’t support the latter two, but we’re working on it (I’m actually wrapping up a small patch to add support for (3)). The cleanest way to do it would be to do this in your `DeltaFilters` class, I think. Anyway, I need to think this through.

lib/delta_filters.rb
stevenharman: WOMP. :frowning: I know it's a reality, but supporting EOL-ed versions of Ruby is no fun. I'll make this change.

stevenharman:
Yes, I agree that this new `DeltaFilters` object is a nice seam for adding in other filters like that. It's not a stretch to add support for both `exclude_types` and `include_types` as well. Rather than taking in an optional array (`exclude_types`), take in an object which can be either an array (in which case we assume they mean `exclude_types`) or a hash. Once it's a hash, we can extract out `exclude_types`, `include_types`, and/or `view` options.

stevenharman:
@khamidou any further changes you'd like to see on this one?

stevenharman:
@khamidou I've rebased this branch atop the latest `master`.

khamidou:
Hi Steven,

Sorry for being late in getting back to you. Unfortunately, I need to think a bit more about how we're going to integrate `include_types` and the other parameters we're planning to add to the delta API.

Karim

stevenharman:
Hello @khamidou,

I've got a rough idea in my head - and I started describing it in an earlier comment.

The gist is, stop adding positional arguments to the `Nylas#delta` (and friends) methods an instead take in an options hash. It could look something like this:

```ruby
def deltas(cursor, options)
  options = [ :exclude_types => options ] unless options.is_a(Hash)
  
  expanded_view = options.delete(:expanded_view) { false }
  filters = build_delta_filters(options)

  # ...
end
```

stevenharman:
Hello @khamidou. I'm nearly to the point that I'll need the `include_types` filter for the Deltas API. If you're happy with the small changes I've made here, I'd like to get them merged and then I can build on top of them to add `include_types`.

lib/inbox.rb
```diff
@@ -90,6 +90,20 @@ class API
...
+      'account' => Inbox::Account,
+      'calendar' => Inbox::Calendar,
+      'contact' => Inbox::Contact,
+      'draft' => Inbox::Draft,
+      'event' => Inbox::Event,
+      'file' => Inbox::File,
+      'folder' => Inbox::Folder,
+      'label' => Inbox::Label,
+      'message' => Inbox::Message,
+      'tag' => Inbox::Tag,
```
zacstewart: I think the tag type is deprecated now.

zspencer:
I really like this PR, but I think it's gotten stale and could be considered superceded by https://github.com/nylas/nylas-ruby/pull/131

I'm going to close this, but if you have objections I can reopen it and integrate it!

PR #39, opened 2015-09-20
Use a Hash to map from status code to exception

stevenharman:


khamidou:
Thanks for the pull request!

PR #42, opened 2015-09-23
Return an external Enumerator when no block given

### LINKED ISSUE #41: Make RestfulModelCollection implement Enumerable
As the `RestfulModelCollection` already implements `#each`, is there any reason not to have it also `include Enumerable`? That would allow consumers to use other handy `Enumerable` methods like `#select` and `#map`.

stevenharman:
Closes #41

### DIFF
--- a/lib/restful_model_collection.rb
+++ b/lib/restful_model_collection.rb
@@ -13,6 +13,8 @@ module Inbox
     end
 
     def each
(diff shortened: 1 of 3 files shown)


khamidou:
Thanks for the pull request, Steven! Merging.

PR #43, opened 2015-09-24
Fix Enuerator documentation and README formatting

stevenharman:
* `#each` without a block returns an `Enumerator`, which can be used like an `Enumerable`. Also, `#select`-ing on a collection of messages will still return a subset of messages. We then map again over that to collect the list of files, which we can download.
* In Ruby, a symbol (e.g., `:files?`) can be converted to a `Proc` by prepending an ampersand (e.g., `&:files #=> :files?.to_proc`). This can be used with methods like `Enumerable#select` and `Enumerable#map` which take a `Proc`/block to filter the enumerable.
* Fix broken code fence blocks.
* Normalize README formatting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,9 +278,9 @@ This allows you to do leverage all that Ruby's `Enumerable` has to offer.
 For example, this is the previous example rewritten to use an `Enumerator`:
 
 ```ruby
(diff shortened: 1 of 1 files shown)


khamidou:
Thanks for fixing this! Merging.

PR #45, opened 2015-10-07
Fix specs broken by addition of `exclude_folders`

stevenharman:
A recent change (052e9146d0ee04d9a5b348c310a4a00e676c69fe) added `exclude_folders=false` to requests on the Delta API, but the specs were not updated to account for the change. This fixes that.


khamidou:
Hgg, you're right. I think what happened is that I fixed the tests locally but forgot to commit them :disappointed:. Anyway, thanks for fixing!

PR #46, opened 2015-10-07
Expose sync_state on APIAccount

stevenharman:
The `/account` API exposes the current `sync_state`, but we were not pulling that across in the `APIAccount` object. Now we do!


khamidou:
Thanks for the pull request! Merging.

PR #47, opened 2015-10-07
Use base Inbox::Error for all errors

stevenharman:
This makes it possible to capture ANY Error raised by Nylas via:

```ruby
def foo
  # do something with `nylas`
rescue Inbox::Error => e
  # handle the error
end
```

rather than needing to know all possible errors which could be raised by Nylas.


khamidou:
Awesome, thanks for fixing!

PR #48, opened 2015-10-09
Update CHANGELOG and add docs about sync_state

stevenharman:


khamidou:
Thanks for the fixes!

PR #49, opened 2015-10-14
Return Enumerator from `#deltas` API when no block is given.

stevenharman:
This behavior is consistent with other APIs which act on a collection.

I also refactored the Delta API specs to user RSpec features and organizational patterns.


stevenharman:
@khamidou Any thoughts on this? I'm finding it very useful for our work.

khamidou:
Thanks for the pull request! Merging.

PR #51, opened 2015-10-21
Change url constructor in url_for_authentication

Bartuz:
You can't pass `state` argument with this method (which should be possible according to your docs). 

Constructing url with string interpolation is not really readable. I believe using URI helpers will make it much more easier and fun to maintain. 

It needs specs probably, I'm not sure if `to_s` casting is necessary  but I'm just giving you idea, maybe I will polish it a bit later.

### DIFF
--- a/lib/inbox.rb
+++ b/lib/inbox.rb
@@ -117,7 +117,7 @@ module Inbox
       arguments = {
         trial: false
       }.merge(options)
-      URI::HTTPS.build(host: @service_domain, query: arguments.to_query)
+      URI::HTTPS.build(host: @service_domain, query: arguments.to_query).to_s
     end
 
     def url_for_management
--- a/lib/inbox.rb
+++ b/lib/inbox.rb
@@ -116,7 +116,11 @@ module Inbox
     def url_for_authentication(redirect_uri, login_hint = '', options = {})
       arguments = {
         trial: false,
-        login_hint: login_hint
+        login_hint: login_hint,
+        client_id: @app_id,
+        response_type: 'code',
+        scpe: 'email',
+        redirect_uri: redirect_uri
       }.merge(options)
       URI::HTTPS.build(host: @service_domain, query: arguments.to_query).to_s
     end
--- a/lib/inbox.rb
+++ b/lib/inbox.rb
@@ -122,7 +122,7 @@ module Inbox
         scpe: 'email',
         redirect_uri: redirect_uri
       }.merge(options)
-      URI::HTTPS.build(host: @service_domain, query: arguments.to_query).to_s
+      URI::HTTPS.build(host: "#{@service_domain}/oauth/authorize", query: arguments.to_query).to_s
     end
 
     def url_for_management
--- a/lib/inbox.rb
+++ b/lib/inbox.rb
@@ -3,7 +3,7 @@ require 'rest-client'
 require 'yajl'
 require 'em-http'
 require 'ostruct'
-
+require 'uri'
 require 'account'
 require 'api_account'
 require 'api_thread'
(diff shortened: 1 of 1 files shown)


lib/inbox.rb
```diff
@@ -114,11 +114,11 @@ def url_for_path(path)
...
-      trialString = 'false'
-      if options[:trial] == true
-        trialString = 'true'
-      end
-      "https://#{@service_domain}/oauth/authorize?client_id=#{@app_id}&trial=#{trialString}&response_type=code&scope=email&login_hint=#{login_hint}&redirect_uri=#{redirect_uri}"
+      arguments = {
+        trial: false,
+        login_hint: login_hint
+      }.merge(options)
+      URI::HTTPS.build(host: @service_domain, query: arguments.to_query).to_s
```
stevenharman: As it is, this won't work as it doesn't include any of the `path` portion, and several of the required `query` arguments are also gone.

That said, I totally agree that [using `URI` or `Addressable` would be much preferred](https://github.com/nylas/nylas-ruby/pull/38#discussion_r39870243) to hand-rolling.

Bartuz: oh sure, totally forgot about default values...

lib/inbox.rb
```diff
@@ -114,11 +114,15 @@ def url_for_path(path)
...
-      "https://#{@service_domain}/oauth/authorize?client_id=#{@app_id}&trial=#{trialString}&response_type=code&scope=email&login_hint=#{login_hint}&redirect_uri=#{redirect_uri}"
+      arguments = {
+        trial: false,
+        login_hint: login_hint,
+        client_id: @app_id,
+        response_type: 'code',
+        scpe: 'email',
+        redirect_uri: redirect_uri
+      }.merge(options)
+      URI::HTTPS.build(host: @service_domain, query: arguments.to_query).to_s
```
stevenharman: Still missing the `oauth/authorize` `path`.

stevenharman:
I agree with the sentiment here, but would prefer to see a holistic approach taken; stop hand-rolling URLs and push that work to either `URL` (and friends) or perhaps `Addressable::URI`.

lib/inbox.rb
Bartuz: This is definitely not my day... :cactus:

Bartuz:
@stevenharman Totally agree. This PR was more like point the issue and show example solution. It needs much more work, because it should be used everywhere in gem. Feel free to close PR and leave it for later reference :)

lib/inbox.rb
stevenharman: :smile:

stevenharman:
@Bartuz I'm a user of the gem and don't have the commit bit. @khamidou works for Nylas and has been the "owner" thus far; it's all up to him.

Bartuz:
@stevenharman so I really appreciate your review here!

lib/inbox.rb
```diff
@@ -114,11 +114,15 @@ def url_for_path(path)
...
-      "https://#{@service_domain}/oauth/authorize?client_id=#{@app_id}&trial=#{trialString}&response_type=code&scope=email&login_hint=#{login_hint}&redirect_uri=#{redirect_uri}"
+      arguments = {
+        trial: false,
+        login_hint: login_hint,
+        client_id: @app_id,
+        response_type: 'code',
+        scpe: 'email',
+        redirect_uri: redirect_uri
+      }.merge(options)
+      URI::HTTPS.build(host: "#{@service_domain}/oauth/authorize", query: arguments.to_query).to_s
```
zacstewart: I may be wrong, but I think `Hash.to_query` comes from activesupport, not native Ruby.

lib/inbox.rb
```diff
@@ -114,11 +114,15 @@ def url_for_path(path)
...
-      if options[:trial] == true
-        trialString = 'true'
-      end
-      "https://#{@service_domain}/oauth/authorize?client_id=#{@app_id}&trial=#{trialString}&response_type=code&scope=email&login_hint=#{login_hint}&redirect_uri=#{redirect_uri}"
+      arguments = {
+        trial: false,
+        login_hint: login_hint,
+        client_id: @app_id,
+        response_type: 'code',
+        scpe: 'email',
```
zacstewart: scope

khamidou:
Bartuz, thanks for taking the time to send a pull request. I ended merging a slightly different [solution](https://github.com/nylas/nylas-ruby/commit/c6617d42780a1deb60f240c7495598b1406edae0). It should be available soon on rubygems.

Karim

PR #64, opened 2015-12-08
Include state param in authentication URL

zacstewart:
* Encodes query parameters using `URI` instead of string interpolation
* Includes state query param as document [here](https://nylas.com/docs/#server_side_explicit_flow)

### DIFF
--- a/lib/inbox.rb
+++ b/lib/inbox.rb
@@ -123,6 +123,9 @@ module Inbox
         'login_hint' => login_hint,
         'redirect_uri' => redirect_uri
       }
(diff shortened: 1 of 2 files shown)


zacstewart:
Jesus. Just realized #51 does exactly this

khamidou:
@zacstewart , thanks for taking sending this pull request! I ended merging a slightly different [solution](https://github.com/nylas/nylas-ruby/commit/c6617d42780a1deb60f240c7495598b1406edae0). It should be available soon on rubygems.

Karim

PR #69, opened 2016-01-20
"state" param for OAuth

icheishvili:
Added state param (from options hash) to url_for_authentication method to help pass contextual information around with OAuth requests.

This is available in the Nylas API here: https://nylas.com/docs/platform?ruby#server_side_explicit_flow but it is not exposed in the ruby gem which is what this PR addresses.

### DIFF
--- a/lib/inbox.rb
+++ b/lib/inbox.rb
@@ -113,11 +113,16 @@ module Inbox
     end
 
     def url_for_authentication(redirect_uri, login_hint = '', options = {})
-      trialString = 'false'
-      if options[:trial] == true
-        trialString = 'true'
-      end
-      "https://#{@service_domain}/oauth/authorize?client_id=#{@app_id}&trial=#{trialString}&response_type=code&scope=email&login_hint=#{login_hint}&redirect_uri=#{redirect_uri}&state=#{options[:state]}"
+      params = {
+        :client_id     => @app_id,
+        :trial         => options[:trial] ? 'true' : 'false',
+        :response_type => 'code',
+        :scope         => 'email',
+        :login_hint    => login_hint,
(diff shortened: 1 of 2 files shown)


lib/inbox.rb
```diff
@@ -117,7 +117,7 @@ def url_for_authentication(redirect_uri, login_hint = '', options = {})
       if options[:trial] == true
         trialString = 'true'
       end
-      "https://#{@service_domain}/oauth/authorize?client_id=#{@app_id}&trial=#{trialString}&response_type=code&scope=email&login_hint=#{login_hint}&redirect_uri=#{redirect_uri}"
+      "https://#{@service_domain}/oauth/authorize?client_id=#{@app_id}&trial=#{trialString}&response_type=code&scope=email&login_hint=#{login_hint}&redirect_uri=#{redirect_uri}&state=#{options[:state]}"
```
khamidou: I think it'd be cleaner if you wouldn't pass `&state` if it's not defined.

khamidou:
Hi @icheishvili, 

Thanks for your pull request. It looks good to me minus the inline comment I added. 

regards,

Karim

lib/inbox.rb
icheishvili: Great idea. How do you feel about me updating some of the other parameters (like login_hint) to also be omitted when they're blank?

khamidou: Sure!

khamidou:
Illia, thanks for taking the time to send a pull request. I ended merging a slightly different [solution](https://github.com/nylas/nylas-ruby/commit/c6617d42780a1deb60f240c7495598b1406edae0). It should be available soon on rubygems.

Karim

Issue #66, opened 2016-01-15
Limit filter doesn't work for collections


khamidou:
For example, this snippet prints subjects for the whole mailbox.

``` ruby
inbox.messages.where(:in => 'Boîte de réception', :limit => 2).each do |msg|
  puts msg.subject
end
```

mechanicles:
We have this issue. How can we add `limit` parameter to `messages` method?

like,

```
nylas_object.messages.where(any_email: email_addresses.join(","), limit: 10)
```

PR #71, opened 2016-01-21
Allow limit and offset filters for collections

DakotaNelson:
The limit and offset filters, which were set correctly in the `where`
function, were being overriden in the `get_model_collection` function
by the offset and limit parameters passed to it (with defaults). The
function now checks if limit or offset are already present in the list
of filters before applying its own.

Also includes new tests of the `where` function.

Close issue #66.

### DIFF
--- a/lib/restful_model_collection.rb
+++ b/lib/restful_model_collection.rb
@@ -15,16 +15,17 @@ module Inbox
     def each
       return enum_for(:each) unless block_given?
 
-      offset = 0
-      chunk_size = 100
+      @filters[:offset] = 0
+      @filters[:limit] = 100
+
       finished = false
       while (!finished) do
-        results = get_model_collection(offset, chunk_size)
+        results = get_model_collection()
         results.each { |item|
           yield item
         }
-        offset += results.length
-        finished = results.length < chunk_size
+        @filters[:offset] += results.length
+        finished = results.length < @filters[:limit]
       end
     end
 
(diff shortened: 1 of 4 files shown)


lib/restful_model_collection.rb
```diff
@@ -138,8 +138,11 @@ def get_model(id)
 
     def get_model_collection(offset = 0, limit = 100)
       filters = @filters.clone
-      filters[:offset] = offset
-      filters[:limit] = limit
+
+      # If filters have already been set for limit or offset, ignore the
```
khamidou: This is pretty confusing. I'd prefer that we handle `limit` and `offset` a level higher (i.e: inside `range`, `each`, etc.)

DakotaNelson: Yeah, agreed. I'll refactor that.

lib/restful_model_collection.rb
```diff
@@ -15,16 +15,17 @@ def initialize(model_class, api, filters = {})
     def each
       return enum_for(:each) unless block_given?
 
-      offset = 0
-      chunk_size = 100
+      @filters[:offset] = 0
```
khamidou: Hmm, I'm not sure this is correct. When doing something like:
```ruby
messages.filter(:subject => 'Hey let's grab some coffee', :filter => 5).each do |msg|
    puts msg.subject
end````

Isn't your code going to get all messages, regardless of whether `filter` and `offset` were set?

File: lib/restful_model_collection.rb [hunk 3]
```diff
@@ -69,7 +70,12 @@ def range(offset = 0, limit = 100)
       end
 
       while (!finished && accumulated.length < limit) do
-        results = get_model_collection(offset + accumulated.length, chunk_size)
+        #results = get_model_collection(offset + accumulated.length, chunk_size)
+        @filters[:offset] = offset + accumulated.length
+        # TODO the below means that if we call range(0, 150) we will make two calls for
```
khamidou: I think it's a pretty simple optimization.

khamidou:
I added a couple comments inline. However, I'm not sure changing `@filters` in place is the right way to go --- I need to think a bit more about it.

lib/restful_model_collection.rb
DakotaNelson: Ah, I think you're right.

I added an `unless @filters.key?(:offset)` for both offset and limit - that way we have defaults, but if something has already been set `each` will respect those parameters.

DakotaNelson:
I added a few more tests to reassure myself, but I agree - having side effects (like setting the filters in place) makes me nervous.

khamidou:
Thanks for fixing!

DakotaNelson:
:+1: Glad to help!

PR #72, opened 2016-01-21
Add `include_types` option to filter deltas

### LINKED ISSUE #68: Add `include_types` for deltas filter
https://www.nylas.com/docs/platform/?curl#filtered_deltas

Now there is only exclude_types

DakotaNelson:
`include_types` is now a valid parameter for both `delta_stream` and
`deltas`. Note that positional parameters for those two functions are
now in a different order after the addition of the `include_types`
parameter.

Closes #68.

### DIFF
--- a/lib/inbox.rb
+++ b/lib/inbox.rb
@@ -258,13 +258,13 @@ module Inbox
     def _build_types_list(types)
       type_string = ""
       types.each do |value|
-        count = 0
         if OBJECTS_TABLE.has_value?(value)
           param_name = OBJECTS_TABLE.key(value)
           type_string += "#{param_name},"
         end
       end
 
+      # remove the dangling comma on the end of the string
       type_string = type_string[0..-2]
     end
 
@@ -279,10 +279,6 @@ module Inbox
         filter_string = _build_exclude_types(exclude_types)
       elsif include_types.any?
         filter_string = _build_include_types(include_types)
-      end
-
-      if include_types.any?
-        include_string = _build_include_types(include_types)
       end
 
       # loop and yield deltas until we've come to the end.
@@ -339,9 +335,7 @@ module Inbox
 
       if exclude_types.any?
         filter_string = _build_exclude_types(exclude_types)
-      end
-
-      if include_types.any?
+      elsif include_types.any?
         filter_string = _build_include_types(include_types)
       end
 
--- a/lib/inbox.rb
+++ b/lib/inbox.rb
@@ -258,13 +258,13 @@ module Inbox
     def _build_types_list(types)
       type_string = ""
       types.each do |value|
+        count = 0
         if OBJECTS_TABLE.has_value?(value)
           param_name = OBJECTS_TABLE.key(value)
           type_string += "#{param_name},"
         end
       end
 
-      # remove the dangling comma on the end of the string
       type_string = type_string[0..-2]
     end
 
@@ -279,6 +279,10 @@ module Inbox
         filter_string = _build_exclude_types(exclude_types)
       elsif include_types.any?
         filter_string = _build_include_types(include_types)
+      end
+
+      if include_types.any?
+        include_string = _build_include_types(include_types)
       end
 
       # loop and yield deltas until we've come to the end.
@@ -326,15 +330,8 @@ module Inbox
       end
     end
 
-    def delta_stream(cursor, options={})
+    def delta_stream(cursor, exclude_types=[], include_types=[], timeout=0, expanded_view=false)
       raise 'Please provide a block for receiving the delta objects' if !block_given?
-
-      raise 'Options argument must be a hash' unless options.is_a? Hash
-
-      exclude_types = options[:exclude_types] || []
-      include_types = options[:include_types] || []
-      timeout = options[:timeout] || 0
-      expanded_view = options[:expanded_view] || false
 
       if exclude_types.any? and include_types.any?
         raise "Cannot pass both include_types and exclude_types parameters"
@@ -342,7 +339,9 @@ module Inbox
 
       if exclude_types.any?
         filter_string = _build_exclude_types(exclude_types)
-      elsif include_types.any?
+      end
+
+      if include_types.any?
         filter_string = _build_include_types(include_types)
       end
 
--- a/lib/inbox.rb
+++ b/lib/inbox.rb
@@ -326,8 +326,15 @@ module Inbox
       end
     end
 
-    def delta_stream(cursor, exclude_types=[], include_types=[], timeout=0, expanded_view=false)
+    def delta_stream(cursor, options={})
       raise 'Please provide a block for receiving the delta objects' if !block_given?
+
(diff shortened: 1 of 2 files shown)


lib/inbox.rb
```diff
@@ -237,32 +237,45 @@ def latest_cursor
...
         end
       end
 
-      exclude_string = exclude_string[0..-2]
+      type_string = type_string[0..-2]
     end
 
-    def deltas(cursor, exclude_types=[], expanded_view=false)
-      return enum_for(:deltas, cursor, exclude_types, expanded_view) unless block_given?
+    def deltas(cursor, exclude_types=[], include_types=[], expanded_view=false)
```
DakotaNelson: Note that this could impact compatibility in existing code.

lib/inbox.rb
```diff
@@ -305,17 +318,23 @@ def deltas(cursor, exclude_types=[], expanded_view=false)
       end
     end
 
-    def delta_stream(cursor, exclude_types=[], timeout=0, expanded_view=false)
+    def delta_stream(cursor, exclude_types=[], include_types=[], timeout=0, expanded_view=false)
```
DakotaNelson: This could also impact existing code.

lib/inbox.rb
```diff
@@ -237,32 +237,45 @@ def latest_cursor
...
         count = 0
         if OBJECTS_TABLE.has_value?(value)
           param_name = OBJECTS_TABLE.key(value)
-          exclude_string += "#{param_name},"
+          type_string += "#{param_name},"
         end
       end
 
-      exclude_string = exclude_string[0..-2]
+      type_string = type_string[0..-2]
```
khamidou: It's not obvious why you're doing this. Could you add a comment?

lib/inbox.rb
```diff
@@ -237,32 +237,45 @@ def latest_cursor
...
+
+    def _build_include_types(types)
+      include_string = "&include_types=" + _build_types_list(types)
+    end
 
-      exclude_types.each do |value|
+    def _build_types_list(types)
+      type_string = ""
+      types.each do |value|
         count = 0
```
khamidou: What's the use of `count` here?

lib/inbox.rb
khamidou: Why would it impact compatibility? Does it make sense to set to `[]` by default?

lib/inbox.rb
```diff
@@ -305,17 +318,23 @@ def deltas(cursor, exclude_types=[], expanded_view=false)
...
+        filter_string = _build_exclude_types(exclude_types)
+      end
+
+      if include_types.any?
+        filter_string = _build_include_types(include_types)
       end
 
       # loop and yield deltas indefinitely.
-      path = self.url_for_path("/delta/streaming?exclude_folders=false&cursor=#{cursor}#{exclude_string}")
+      path = self.url_for_path("/delta/streaming?exclude_folders=false&cursor=#{cursor}#{filter_string}")
```
khamidou: Why pass only `filter_string` to the URL?

lib/inbox.rb
DakotaNelson: Very good question - I had completely ignored it. It looks like it must be left over from some old code. I just removed it.

lib/inbox.rb
DakotaNelson: Done!

lib/inbox.rb
DakotaNelson: My thinking is that if someone has set `expanded view` in their code (as a positional argument), it's now in the wrong order.

I'm using an empty array here as a sentinel, more than anything - see a few lines down where it tests to see if there's anything in the array (`if include_types.any?`), and ignores the parameter if not.

lib/inbox.rb
DakotaNelson: Since we can only have either an `include_types` or `exclude_types` parameter, and never both, `filter string` is either of the form `&include_types=a,b,c` or `&exclude_types=a,b,c`. (or it's empty, if neither is passed)

lib/inbox.rb
khamidou: Hmm, If I recall correctly, Ruby does weird thing with positional arguments (i.e: https://eval.in/506574). It would be pretty useful if you could either:
- figure out how to not break backward compatibility
- otherwise refactor the function by using a parameter hash.

lib/inbox.rb
khamidou: Okay then make it explicit by having an if/else block instead of two consecutive if blocks in L328-332.

DakotaNelson: Yep, that's already in place as of 34221a09a1629affc837c6c820c748b8c958f486.

lib/inbox.rb
DakotaNelson: The way I had originally considered was to make the function footprint be `deltas(cursor, exclude_types=[], expanded_view=false, include_types=[])`, which preserves compatibility but is a little weird.

Thoughts on that vs. refactor into a parameter hash? I worry that will break compatibility anyway, but it'll be less weird and make things easier in the future if more parameters are added.

DakotaNelson: If we're willing to break compatibility with Ruby < 2, we can refactor to have the function footprint be:

`deltas(cursor, exclude_types: [], include_types: [], expanded_view: [])`

Which gives us great backward compatibility in the future, at the cost of breaking it now for a subset of people (anyone calling `deltas` with any parameter other than just `cursor` will have issues) and also losing support for older ruby.

lib/inbox.rb
modernptz: I think better to use hash as parameter here e.g.:  {include: :messages, exclude: :files, :threads, expanded_view: true}

lib/inbox.rb
khamidou: Okay, let's do this!

khamidou:
@DakotaNelson can you address my last comment so I can merge this?

lib/inbox.rb
stevenharman: Note... the `nylas` Ruby Gem v2.0 was release the other day and this change was NOT in there. If this is merged, we'll want to bump to v3.0 to indicate the breaking change per SemVer.

stevenharman: Also, as I'd suggested in #38, using keyword args would make this far cleaner, and allow for a non-breaking API going forward.

If we're going to have a breaking change, I think we should move to keyword args across the board - they allow a more flexibility in future changes, while actually reducing coupling. Specifically, keyword arguments are an example of [Connascence of Name, which is preferable to Connascence of Position](https://en.wikipedia.org/wiki/Connascence_(computer_programming)).

zspencer:
Thanks @DakotaNelson ! I may use this when we start resolving #68 in Version 3.3.

zspencer:
Deprecated by https://github.com/nylas/nylas-ruby/pull/131

I would be in favor of either a different class that we could use to prototype the keyword argument versions of the CLI and give an upgrade path that we could deprecate when moving to `4.0`

PR #73, opened 2016-01-22
Refactor `delta_stream` to allow multiple streams

DakotaNelson:
`delta_stream` is a blocking call, and previously had no good way to run
in parallel in order to stream several accounts at once.

EventMachine's run loop was removed from the library so that SDK
implementers can create multiple streams inside their own EventMachine
run block.

`delta_stream` must be wrapped in an `EventMachine.run` block in order
to function, whereas before it could be called plain. Two tests have
been refactored to reflect this, and documentation has been updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,23 +458,41 @@ The streaming API will receive deltas in real time, without needing to repeatedl
 ```ruby
 cursor = nylas.latest_cursor
 
-last_cursor = nil
-nylas.delta_stream(cursor) do |event, object|
-  if event == "create" or event == "modify"
-    if object.is_a?(Nylas::Contact)
-      puts "#{object.name} - #{object.email}"
-    elsif object.is_a?(Nylas::Event)
-      puts "Event!"
+EventMachine.run do
+  nylas.delta_stream(cursor) do |event, object|
+    if event == "create" or event == "modify"
+      if object.is_a?(Nylas::Contact)
+        puts "#{object.name} - #{object.email}"
+      elsif object.is_a?(Nylas::Event)
+        puts "Event!"
+      end
+    elsif event == "delete"
+      # In the case of a deletion, the API only returns the ID of the object.
+      # In this case, the Ruby SDK returns a dummy object with only the id field
+      # set.
+      puts "Deleting from collection #{object.class.name}, id: #{object}"
     end
-  elsif event == "delete"
-    # In the case of a deletion, the API only returns the ID of the object.
-    # In this case, the Ruby SDK returns a dummy object with only the id field
-    # set.
-    puts "Deleting from collection #{object.class.name}, id: #{object}"
   end
-  last_cursor = object.cursor
+end
+```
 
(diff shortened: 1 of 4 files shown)


lib/inbox.rb
```diff
@@ -347,15 +347,13 @@ def delta_stream(cursor, exclude_types=[], timeout=0, expanded_view=false)
         end
       end
 
-      EventMachine.run do
-        http = EventMachine::HttpRequest.new(path, :connect_timeout => 0, :inactivity_timeout => timeout).get(:keepalive => true)
-        http.stream do |chunk|
-          parser << chunk
-        end
-        http.errback do
```
DakotaNelson: I hate to remove this. I'll try to find a way to put it back in.

DakotaNelson: Done - see below.

README.md
```diff
@@ -458,23 +458,41 @@ The streaming API will receive deltas in real time, without needing to repeatedl
...
-    # In the case of a deletion, the API only returns the ID of the object.
-    # In this case, the Ruby SDK returns a dummy object with only the id field
-    # set.
-    puts "Deleting from collection #{object.class.name}, id: #{object}"
   end
-  last_cursor = object.cursor
+end
+```
+
+In order to receive streams from multiple accounts, simply create multiple API handles (one for each account), then open a `delta_stream` for each of them.
```
khamidou: This isn't very clear. Maybe heva seomthing like "To receive streams from multiple accounts, call `delta_stream` for each of them inside an `EventMachine.run` block.

README.md
```diff
@@ -458,23 +458,41 @@ The streaming API will receive deltas in real time, without needing to repeatedl
...
 
-  # This will loop indefintely
+```ruby
+api_handles = []
+nylas.accounts.each {
```
khamidou: I think you can do without creating the handles. Most of our customers are using the hosted API, so this part will be different for them anyway. Simply have a line like:
```ruby
api_handles = [] # a list of Nylas::API objects.
```

lib/inbox.rb
khamidou: Hmm, not sure I'm following.

khamidou:
Generally LGTM, minus the comments I added inline. Also, since this is a breaking change, I'd like you to edit CHANGELOG.md to add this change under `Upcoming release`.

README.md
DakotaNelson: Sounds good - done!

README.md
DakotaNelson: Ah, good catch. I'll change that.

lib/inbox.rb
DakotaNelson: Ah, I meant the `http.errback` functionality to catch errors - I removed, then re-added it.

DakotaNelson:
:+1:

khamidou:
LGTM, thanks for fixing!

PR #74, opened 2016-02-16
Fix errors pointed out in 849c11e

DakotaNelson:
Also includes several new tests


lib/restful_model_collection.rb
```diff
@@ -37,6 +37,9 @@ def count
     end
 
     def first
+      @filters[:offset] = 0
```
khamidou: I'm not sure this is solving the completely. Isn't the root cause that the filters were shared by all subclasses of `RestfulModelCollection`?

DakotaNelson: I just added a test (https://github.com/nylas/nylas-ruby/pull/74/files#diff-ebbc5b209198fb9bc40897cc756c8262R212) to make sure that isn't the case, and it seems like we're good.

zspencer:
I believe this was resolved by #112?

Issue #44, opened 2015-10-06
JRuby support


jockee:
The `yajl-ruby` gem is not compatible. Looks like https://github.com/dgraham/json-stream is.

PR #76, opened 2016-03-13
JRuby Support

mguymon:
I was able to add JRuby Support by creating a [simple stream parser in Java](https://github.com/mguymon/simple_stream) based on [JSON Simple](https://github.com/fangyidong/json-simple) to replace YAGJL and its C dependencies. The nuts and bolts of loading java jars is tied up neatly with a bow using [lock_jar](https://github.com/mguymon/lock_jar). 

I created platform dependent mixins as `Nylas::Inbox::Java` and `Nylas::Inbox::Ruby`. These are mixed in to the `Inbox::API` at load time and prevent platform dependent concerns from leaking.

~~Unfortunately, Eventmachine was tossing a NullPointer error. Not sure how up to date their java implementation is. I ended up using [httprd](https://github.com/httprb/http) instead, since async is less of an issue with JRuby.~~ Replaced with a java impl as part of https://github.com/mguymon/simple_stream

While the tests pass in JRuby and Ruby, I would consider this still a WIP. I plan on testing the JRuby impl on my local running sync-engine to make sure it is parsing the JSON correctly.

----
 Outstanding Tasks
- [x] Add support for deploying platform specific gems
- [x] Update README
- [x] Get Travis to play nice with JRuby
- [x] Integration testing with running sync-engine


mguymon:
Should make https://github.com/nylas/nylas-ruby/issues/44 happy

mguymon:
Seem to be hitting travis being unable to install Ruby 2.3.0 on OSX https://github.com/travis-ci/travis-ci/issues/5361

mguymon:
And Java 1.8 does not work on OSX from Travis, quite the messy I have stumbled into - https://github.com/travis-ci/travis-ci/issues/4957

mguymon:
So after a bit of a battle with Travis, I got JRuby 9k passing. Unfortunately the older JRuby 1.7 failed at installing the event machine gem so I left it off. I wonder if the older gem is not compat with Java 1.8.

Gemfile
```diff
@@ -14,7 +16,16 @@ group :development, :test do
...
   gem 'pry-nav'
-  gem 'pry-stack_explorer'
+  gem 'pry-stack_explorer', platform: :ruby
   gem 'webmock'
   gem 'sinatra'
 end
+
+@@check ||= at_exit do
+  begin
+    require 'lock_jar/bundler'
```
khamidou: Hmm, won't this fail for MRI?

mguymon: It will, but it is wrapped in `rescue LoadError` to silently fail when it cannot find the `lock_jar` gem in the Bundle.

This block is a helper to regenerate the Jarfile.lock when bundler is called, removing the additional step of having to lock it separately.

lib/inbox.rb
```diff
@@ -84,6 +83,14 @@ def self.interpret_response(result, result_content, options = {})
 
 
   class API
+    if RUBY_PLATFORM[/java/] == 'java'
+      require 'nylas/inbox/java'
+      include Nylas::Inbox::Java
```
khamidou: I like the idea of having platform-specific modules, but I'd prefer another name here. What about `Nylas::Java`?

mguymon: :+1:

khamidou:
This PR looks good to me, minus the comments I left.

Gemfile
khamidou: Hmm, can you add an explicit check for JRuby and a one-line comment explaining what it does? It would make the intent of the code clearer.

mguymon: Sounds good, this is definitely the ugly side of the jar loading.

mguymon:
Got a simple prototype working using [Vertx.io](http://vertx.io/) and the Nylas. Successfully fetched a few emails over the api using JRuby. Next step is to tie into the Vertx stream handler and put the api through its paces.

mguymon:
The `httprd` was failing to stream correctly, so I ended up handling it in [Java instead](https://github.com/mguymon/simple_stream/blob/master/src/main/java/com/tobedevoured/json/SimpleStream.java#L125-L169). After coercing the results into Ruby field collections, everything worked correctly on my local sync:

     Executing callback with {"cursor":"<cursor>","attributes":{"cc":[],"snippet":"<snippet goes here>","labels":[],"thread_id":"<thread_id>","account_id":"<account_id>","reply_to":[],"starred":false,"files":[],"from":[{"name":"Michael Guymon","email":"<emai>l"}],"to":[{"name":"Michael N Guymon","email":"<emai>l"}],"id":"id","events":[],"object":"message"},"id":"<id>","event":"modify","object":"message"}

mguymon:
I am pumped, I got the Nylas API to work with [Vertx's Event Bus](http://vertx.io/docs/vertx-core/ruby/#_the_event_bus_api) to handle the streaming and parsing of the delta stream. This means I can write other consumers (in different languages if need be) that operate on the delta stream feed. 

The skunk work is in https://github.com/mguymon/mailnipulator but there is not much to see at the moment.

Ok, so back to the task on hand. I will finish up the remaining work of  creating a script to release separate MRI and JRuby gems and update the README with some details about the JRuby impl.

khamidou:
Awesome! I'll review this tomorrow!

.travis.yml
```diff
@@ -6,5 +8,15 @@ rvm:
...
+  - "osx"
+
+matrix:
+  allow_failures:
+  # https://github.com/travis-ci/travis-ci/issues/5361
+  - os: osx
+    rvm: 2.3.0
+  include:
+  - os: linux
+    rvm: jruby-9.0.4.0
```
khamidou: Awesome!

Gemfile
```diff
@@ -14,7 +17,16 @@ group :development, :test do
...
-  gem 'pry-stack_explorer'
+  gem 'pry-stack_explorer', platform: :ruby
   gem 'webmock'
   gem 'sinatra'
 end
+
+@@check ||= at_exit do
+  # JRuby only.Generate Jarfile.lock which is used by LockJar to load jar
+  # dependencies into the classpath.
+  if RUBY_PLATFORM[/java/] == 'java'
```
khamidou: Thanks for fixing this.

Rakefile
```diff
@@ -50,9 +50,22 @@ end
 task :nylas_build do
   Jeweler::Tasks.new do |gem|
     setup_nylas_gem(gem)
+    gem.platform = 'java'
```
khamidou: I've never published a JRuby gem before --- Does JRuby use a different rubygems repository? Also, won't your changes prevent us from publishing the MRI version of the gem?

lib/inbox.rb
```diff
@@ -84,7 +83,7 @@ def self.interpret_response(result, result_content, options = {})
 
 
   class API
-    attr_accessor :api_server
+    attr_accessor :api_server, :stream_handler
```
khamidou: For consistency reasons, can you move the `:stream_handler` declaration to the next line?

e.g: 
```ruby
attr_accessor :api_server
attr_accessor :stream_handler
```

khamidou:
Your PR looks good to me, minus the inline comments I added.

Rakefile
mguymon: Yeah, I did not mean to commit that. Still need to finish up the work for publishing separate gems. 

Publishing a JRuby gem is the same as MRI, the difference is you need to set the platform. The work flow wrinkle is you have to build the gem while in the correct Ruby,e.g. rbenv switch to MRI than switch to JRuby to publish both gems. This can be wrapped up in a shell script.

mguymon:
@khamidou Got distracted by some other work, but finally got everything wrapped up.

I mucked with the Rakefile and Jeweler so it will create gems with the correct platform dependencies. I created a release script [bin/release](https://github.com/mguymon/nylas-ruby/blob/jruby_support/bin/release) that switches between MRI 2.2.2 and JRuby 9.0.4.0 for releasing the Nylas and Inbox gem. . . *but* it uses `rbenv` to do the switching, so will have to update it if you use something else to handle the ruby versions.

mguymon:
I went and made a separate gem for handling the json streaming in JRuby, [sjs]( https://github.com/mguymon/sjs). It felt ugly having all the java specific stuff in the stream handler. I also did a could more revision on the base java code to buffer and handle partial fragments correctly.

@khamidou how does it look now? I believe there is still an outstanding concern of the best way to release 2 gems.

README.md
```diff
@@ -493,6 +501,43 @@ EventMachine.run do
 end
 ```
 
+#### JRuby
+
+JRuby implementation uses [Simple Stream](https://github.com/mguymon/simple_stream) instead of
```
khamidou: `The JRuby implementation`

README.md
```diff
@@ -493,6 +501,43 @@ EventMachine.run do
...
+EventMachine and YAJL. No need for the `EventMachhine.run` block.
```
khamidou: Typo --- `EventMachhine.run`

khamidou:
Apart from my inline comments, this looks great to me!

mguymon:
@khamidou Fixed the README, I think we are good to go :fireworks:

PR #91, opened 2016-12-09
Add ability to get collection of expanded messages

jastkand:
Sometimes you need to get collection of expanded messages and it's not efficient to do

```ruby
api.messages.each do |message|
  expanded_message = message.expanded
  puts expanded_message.message_id
end
```

because it produces an additional query to the Nylas API for each `expanded` method call.

This PR adds ability to request the collection of expanded messages, like

```ruby
api.messages(expanded: true).each do |expanded_message|
  puts expanded_message.message_id
end
```

or 

```ruby
thread.messages(expanded: true).each do |expanded_message|
  puts expanded_message.message_id
end
```

It also allows getting single expanded message in one API call with

```ruby
expanded_message = api.messages(expanded: true).find(id)
```

which is more efficient than getting it with two queries

```ruby
message = api.messages.find(id) # gets a message
expanded_message = message.expended # gets an expanded message
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ gem install nylas
 
 ### Sinatra App
 
-A small example of a Sintra app is included in the `examples/sinatra` directory. You can check-out the `README.md` in the sinatra folder to learn more about the example
+A small example of a Sinatra app is included in the `examples/sinatra` directory. You can check-out the `README.md` in the sinatra folder to learn more about the example
 
 ```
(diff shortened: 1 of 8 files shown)


pfista:
Thanks for taking the time to submit this PR!

rexmadden:
Is this going to be released soon? It would save a lot of bandwidth!

PR #97, opened 2017-03-13
Update 'rest-client' to v2

mechanicles:
Actually, in our Rails project, we want to update `mailgun-ruby` gem but newer version of this gem is dependent on the `rest-client` `~> 2.0` but `nylas-ruby` gem is dependent on `rest-client` `~> 1.6` and this causing an issue. It would be good if we merge this as soon as possible.

Let me know your thoughts.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'rest-client', '~> 1.6'
+gem 'rest-client', '>= 1.6'
 gem 'yajl-ruby', '~> 1.2', '>= 1.2.1', platform: :ruby
 gem 'em-http-request', '~> 1.1', '>= 1.1.3', platform: :ruby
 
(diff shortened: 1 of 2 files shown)


mechanicles:
@pfista Any update on this?

scollon42:
We strongly need this update too. If someone can review and merge this pull request It would be awesome ! :+1:

jastkand:
@khamidou @pfista @stevenharman could you reveiw this, please?

Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'rest-client', '~> 1.6'
+gem 'rest-client', '~> 2.0'
```
jastkand: @mechanicles i think it's better to put the `>= 1.6` here. It will allow to use both 1.6, 1.8 versions (if someone needs them) and 2.0.

mechanicles:
@jastkand I have updated code. Now it supports versions `>= 1.6`. Please check.

zspencer:
I believe this is also resolved by #111

jastkand:
@mechanicles yes, this can be closed

mechanicles:
Thanks, @zspencer @jastkand. Great Job!

Issue #104, opened 2017-07-13
Ruby 2.4+ - the version of rest-client specified in the Gemfile is too old, and throws errors.


xxx:
It runs into https://github.com/rest-client/rest-client/issues/612 , which is resolved in newer versions of rest-client.

rexmadden:
Also having the same problem.

jwg2s:
Also experiencing this in trying to evaluate whether nylas will work for us

Issue #81, opened 2016-07-04
2.0.1 undefined method `[]' for #<Set: {#<MIME::Type: application/json>}>


g3ortega:
Hi, I'm having an issue when I try to send a draft, after updating from 2.0 to 2.0.1, using ruby 2.2.4

Here a test with the same snippet you provide. 

``` ruby
2.2.4 :018 > nylas.drafts.build({ :to => [{:name => 'Ben Gotow',
2.2.4 :019 >                                         :email => 'algo@algo.com'}],
2.2.4 :020 >                                     :subject => "Sent by Ruby",
2.2.4 :021 >                                     :body => "Hi there!<strong>This is HTML</strong>"}).send!
NoMethodError: undefined method `[]' for #<Set: {#<MIME::Type: application/json>}>
        from /Users/gerardo/.rvm/gems/ruby-2.2.4/gems/rest-client-1.6.7/lib/restclient/request.rb:307:in `type_for_extension'
        from /Users/gerardo/.rvm/gems/ruby-2.2.4/gems/rest-client-1.6.7/lib/restclient/request.rb:312:in `type_for_extension'
        from /Users/gerardo/.rvm/gems/ruby-2.2.4/gems/rest-client-1.6.7/lib/restclient/request.rb:278:in `block in stringify_headers'
        from /Users/gerardo/.rvm/gems/ruby-2.2.4/gems/rest-client-1.6.7/lib/restclient/request.rb:272:in `each'
        from /Users/gerardo/.rvm/gems/ruby-2.2.4/gems/rest-client-1.6.7/lib/restclient/request.rb:272:in `inject'
        from /Users/gerardo/.rvm/gems/ruby-2.2.4/gems/rest-client-1.6.7/lib/restclient/request.rb:272:in `stringify_headers'
        from /Users/gerardo/.rvm/gems/ruby-2.2.4/gems/rest-client-1.6.7/lib/restclient/request.rb:92:in `make_headers'
        from /Users/gerardo/.rvm/gems/ruby-2.2.4/gems/rest-client-1.6.7/lib/restclient/request.rb:58:in `initialize'
        from /Users/gerardo/.rvm/gems/ruby-2.2.4/gems/rest-client-1.6.7/lib/restclient/request.rb:33:in `new'
        from /Users/gerardo/.rvm/gems/ruby-2.2.4/gems/rest-client-1.6.7/lib/restclient/request.rb:33:in `execute'
[...]

PR #111, opened 2017-11-07
Add support for rest-client 1.6>2.0>

zspencer:
The nylas-ruby usage of rest-client is not really tied to the particular feature sets in sub 2.0 rest-client, but some of the tests were picky. 

I've made the tests less greedy while maintaining their utility.

Resolves https://github.com/nylas/nylas-ruby/issues/104
Resolves https://github.com/nylas/nylas-ruby/issues/81


zspencer:
Confirmed that this also resolves https://github.com/nylas/nylas-ruby/issues/81 (though theoretically, updating their rest-client to 1.7 or 1.8 would resolve the issue as well)

spec/restful_model_spec.rb
```diff
@@ -58,12 +58,14 @@ def foo=(bar); end
 
     it "should pass parameters as query parameters when calling delete" do
       url = 'http://localhost:5555/events/1'
-      stubbed_url = 'http://localhost:5555/events/1?param1=&param2=stuff&send_notifications=true'
+      stubbed_url = /events\/1\?param1[=]?&param2=stuff&send_notifications=true/
```
khamidou: Hmm, could you give me a bit more details about the change you made to the test?

khamidou:
LGTM, I just have one tiny question about the changes you made to the unit test.

spec/restful_model_spec.rb
zspencer: Sure!

WebMock supports stubbing URLs in a number of ways, ranging from a straight string comparison, to a "with" syntax that describes the different parts of the URI, to a regular expression.

From my perspective, tests should focus on the thing they actually care about, in this case "Hey, when you send a parameter as nil, there isn't a value included in the query string" and not the incidentals ("Hey, we send all our requests to the right host, right?") that are taken care of in other parts.

I removed the bit of the stub/assertion that expects localhost:5555 (since that is incidental) and switched to regex syntax so that the library change where rest-client went from sending `param=` to `param` for nil params doesn't impact the test anymore.

My "ideal" API tests follow the arrange/act/assert format:

```
# arrange
# (Make stubs as un-greedy as possible)

# act
# fire off the actual request

# assert
# Make asserts about what query parameters got what value in the stubs here, and make asserts against the resulting object
```

I, personally, don't mind multiple asserts per test definition;  though if you have strong feels about that we'd probably want to look into something like `rspec-given` to make the Arrange/Act/Assert pattern more explicit while exposing each assert as it's own test definition.

I didn't want to make *too* broad of a change in this PR, however.

khamidou: Make sense, thanks!

PR #112, opened 2017-11-08
Each and range no longer break limit filters

### LINKED ISSUE #66: Limit filter doesn't work for collections

zspencer:
I want to meditate on preserving the "evaluate each page as it's loaded" behavior overnight. Will take another stab at it in the morning

Resolves #66

### DIFF
--- a/lib/restful_model_collection.rb
+++ b/lib/restful_model_collection.rb
@@ -15,8 +15,10 @@ module Nylas
     def each
       return enum_for(:each) unless block_given?
 
-      get_model_collection.each do |item|
-        yield item
+      get_model_collection do |items|
+        items.each do |item|
+          yield item
+        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


khamidou:
Hi @zspencer! If I understand correctly, you moved the limiting/offsetting logic from `range` to `get_model_collection`?

khamidou:
Hi @zspencer! If I understand correctly, you moved the limiting/offsetting logic from `range` to `get_model_collection`?

zspencer:
@khamidou - Yep, basically pulled that logic up out of range and into get_model_collection, and further refined it to support the use-cases of `#each`

khamidou:
Talked offline – thanks for fixing this!

Issue #94, opened 2017-01-27
Update in drafts - File_ids getting overwritten


bhattjay:
If your current version of draft already has some files uploaded and is saved, and now if you first of all find this draft, and then make a change in any other field, then this new version overwrites the existing file_ids of the existing draft. It passes over an empty array in the update, and hence, you lose out on the already saved files.

zspencer:
@bhattjay - This has been resolved in 4.0, as we have added an `update` method which only sends the change data instead of all the attributes. Save may still exhibit this behavior, but I would hope we can add dirty-checking at some point.

If the problem persists beyond once upgrading to 4.X please re-open this ticket.

PR #113, opened 2017-11-08
Only send data to the API when the data has changed on model update/save calls

zspencer:
Resolves https://github.com/nylas/nylas-ruby/issues/94


zspencer:
This is going to fail horribly, but I wanted to get it up before clocking out for the day. Most of the failures are the result of over-specification in the stubs, but I want to make sure we're not breaking backwards compatibility before merging.

in particular, I need to make sure there aren't any required parameters that I'm not aware of that would be excluded using this logic.

zspencer:
No longer necessary, as the old structures are being deprecated. May be worth merging into the 3.X branch if people really want the bug fix.

PR #131, opened 2017-11-14
Add support for include_types to delta stream and deltas

zspencer:


zspencer:
The interface for this is a little wonky and I'm not sure I like it, but I also didn't want to muck about with adding in keyword arguments.

Part of me would like to sprout a new "API" class that can have nicer syntactical sugar, but I'd rather get the standard API Class playing nicely and then we can talk about prototyping a new interface as part of a later milestone.

khamidou:
LGTM, thanks for making this cleaner!

Issue #118, opened 2017-11-09
Cast Message and Draft's `date` field to an actual date object instead of string


zspencer:
Messages and Drafts have a `date` field, which is a timestamp but is parsed in the ruby SDK as a string.

We'll want to do this in a backwards compatible way, perhaps by adding a writer/reader method that does the casting and updates the actual parameter?

PR #133, opened 2017-11-14
Expose date fields as ruby dates

zspencer:
Resolves https://github.com/nylas/nylas-ruby/issues/118

I chose to not overwrite the existing `date` method as that could result in surprises for our consumers. Instead, I went with overwriting the `date=` so that it coerces the date to a string when it's set with a real date object, and then it parses it back out as `parsed_date` for convenience.

I'll need to update this when we merge #113

### DIFF
--- a/lib/nylas/message.rb
+++ b/lib/nylas/message.rb
@@ -45,6 +45,30 @@ module Nylas
       end
     end
 
+
+    # Convenience method to access a Ruby Date version of the `date` parameter
+    # Will replace the string version of`#date` when we move to 4.0
+    def parsed_date
+      if date.kind_of?(Numeric)
+        Time.at(date).to_datetime
+      else
+        DateTime.parse(date)
+      end
+    end
+
+    alias_method :orig_date=, :date=
+    # Ensures when we are setting the date on this instance
+    # that we are storing it to a unix timestamp.
+    def date=date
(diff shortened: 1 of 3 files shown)


zspencer:
@khamidou - I've updated this based on your feedback.

Something to note, at least in API V1, the "date" field on the drafts and messages endpoints are unix epoch timestamps, not the objects you pointed me towards on the event API.

That said, I really like the way you represent your kinds of dates in the event API and will want to make sure we are keeping them in mind.

zspencer:
added readme demonstrating the events.

khamidou:
This looks good to me but I have two questions:
1. The thread object also has a couple date-like object like `first_message_timestamp`, `last_message_received_timestamp`, etc. Are you planning to support these in a separate diff?
2. Where are the changes to the README you said you were adding in https://github.com/nylas/nylas-ruby/pull/133#issuecomment-347727816

Regards,

zspencer:
@khamidou - The events readme changes were in the message/draft events PR, I left this comment on the wrong PR: https://github.com/nylas/nylas-ruby/pull/135/files

The timestamps you're discussing were already merged: https://github.com/nylas/nylas-ruby/blob/master/lib/nylas/thread.rb#L23 and those already are `time_attr_accessor`-ified.

Issue #57, opened 2015-11-08
Moving message to folder


g3ortega:
i was trying to move a message using the sample in the docs:

``` ruby
trash_id = 'ds36ik7o55gdqlvpbrjbg9ovn'
message.folder = trash_id
message.save!
```

And i got this exception:

```
NoMethodError: undefined method `id' for "b8n1of68aixksz3o84qstro7i":String
from /Users/sample/.rvm/gems/ruby-2.2.0/gems/nylas-1.1.0/lib/message.rb:64:in `as_json'
```

I was reviewing the code and the problem originates in this method:

``` ruby

    def as_json(options = {})
      hash = {}

      # unread, starred and labels/folder are the only attribute
      # you can modify.
      if not @unread.nil?
        hash["unread"] = @unread
      end

      if not @starred.nil?
        hash["starred"] = @starred
      end

      if not @labels.nil? and @labels != []
        hash["labels"] = @labels.map do |label|
          label.id
        end
      end

      if not @folder.nil?
        hash["folder"] = @folder.id
      end

      hash
    end
```

Apparently an object of type Thread is expected, because the id is asked. I was also trying with labels and I have the same issue. On the other hand, i think the use of 'unless' instead of 'if not' would be more readable in a ruby context :smile: 

I can help with this issue, if it's actually is one, and not an error in the docs, and also I can help with some refactoring and optimizations.

zspencer:
Yeah, the interface for setting a folder and / or label is a bit confusing.

Right now, we expect  `folder` to be an instance of `Nylas::Folder`, however we don't do any casting or coercing into that type. 

I think it would be reasonable to update the examples to make that explicit.

PR #134, opened 2017-11-14
Make type errors for message folders and labels explicit

zspencer:
This resolves https://github.com/nylas/nylas-ruby/issues/57 by making the interface more explicit about the types it expects.

I generally prefer to do this on setting the data as opposed to on parsing, what do you think @khamidou? An example of doing the casting/validation on input is here: https://github.com/nylas/nylas-ruby/pull/133

### DIFF
--- a/lib/nylas/message.rb
+++ b/lib/nylas/message.rb
@@ -28,7 +28,7 @@ module Nylas
       @cc ||= []
       @bcc ||= []
       @labels ||= []
-      @folder ||= nil
+      self.folder ||= nil
 
       # This is a special case --- we receive label data from the API
       # as JSON but we want it to behave like an API object.
@@ -60,12 +60,19 @@ module Nylas
 
       if not @labels.nil? and @labels != []
         hash["label_ids"] = @labels.map do |label|
+          if !label.respond_to?(:id)
+            raise TypeError, "label #{label} does not respond to #id"
+          end
           label.id
         end
       end
 
-      if not @folder.nil?
-        hash["folder_id"] = @folder.id
+      if !folder.nil? && !folder.respond_to?(:id)
+        raise TypeError, "folder #{folder} does not respond to #id"
+      end
+
+      if !folder.nil? && folder.respond_to?(:id)
+        hash["folder_id"] = folder.id
       end
 
       hash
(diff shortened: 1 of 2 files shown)


lib/message.rb
```diff
@@ -60,12 +60,19 @@ def as_json(options = {})
...
 
-      if not @folder.nil?
-        hash["folder_id"] = @folder.id
+      if !folder.nil? && !folder.respond_to?(:id)
+        raise TypeError, "folder #{folder} does not respond to #id"
+      end
+
+      if !folder.nil? && folder.respond_to?(:id)
+        hash["folder_id"] = folder.id
       end
```
zspencer: We may want to pomote this to `ModelState#as_json` later on.

lib/message.rb
```diff
@@ -28,7 +28,7 @@ def inflate(json)
       @cc ||= []
       @bcc ||= []
       @labels ||= []
-      @folder ||= nil
+      self.folder ||= nil
```
khamidou: curious, why the `self` here?

lib/message.rb
khamidou: Hmm, do you mean `RestfulModel`?

zspencer: Ah, no. model_state is a class that's job is to track whether or not the data has mutated. It's part fo the branch that we will be merging (hopefully soon) to ensure we only send the changed data when posting to the API to reduce the prevalance of "overwrite" bugs.

lib/message.rb
zspencer: Writing directly to instance variables in Ruby means there's no seam to inject behavior, so I tend to prefer using attribute writers and readers. This way if we need to do something like say, rename the instance variable, we can leave the old attr reader as an alias and seamlessly upgrade.

Basically it makes it slightly easier to make changes without making changes across the file.

The reason for `self` here is because Ruby thinks that if you do `folder =` you are assigning a local variable, as opposed to using the setter method. Hence `self.folder =`. Python gets around this with an explicit `self` which  (IMO) despite being slightly more verbose is also more user-friendly.

lib/message.rb
zspencer: Here's the sprouted RestfulModel class: https://github.com/nylas/nylas-ruby/pull/113/files#diff-a351f8985b331720cad0220e410159beR9

Issue #119, opened 2017-11-09
Support reading `events` fields for Messages and Drafts API


zspencer:
Drafts and messages all support `events`, but the ruby-sdk does not at this time. Add support for linking a message or draft to an event: per https://docs.nylas.com/reference#get-drafts and https://docs.nylas.com/reference#events

zspencer:
Reading through the API docs, it looks like *writing* event_ids isn't supported by the messages and draft apis. @khamidou or @pfista could you confirm this? Or is the API feature undocumented?

PR #135, opened 2017-11-14
Message and Draft objects expose their related events

zspencer:
Resolves https://github.com/nylas/nylas-ruby/issues/119

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,6 +239,7 @@ thread.save!
 # List messages
 thread.messages.each do |message|
   puts message.subject
+  puts message.events
 end
 
 # List expanded messages (with Message-Id, In-Reply-To and References fields)
@@ -323,6 +324,7 @@ If you've building your own threading solution, you'll probably need access to a
 
 ```ruby
 msg = nylas.messages.first
+puts msg.events
 expanded_message = msg.expanded
 puts expanded_message.message_id
 puts expanded_message.references
--- a/lib/nylas/event.rb
+++ b/lib/nylas/event.rb
@@ -41,7 +41,5 @@ module Nylas
 
       self
     end
-
-
   end
 end
--- a/lib/nylas/message.rb
+++ b/lib/nylas/message.rb
@@ -5,6 +5,8 @@ require 'nylas/mixins'
 module Nylas
   class Message < RestfulModel
 
+    attr_reader :events
+
     parameter :subject
     parameter :snippet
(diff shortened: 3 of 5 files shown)


lib/restful_model_collection.rb
```diff
@@ -3,7 +3,7 @@
 module Nylas
   class RestfulModelCollection
 
-    attr_accessor :filters
+    attr_accessor :filters, :_api, :model_class
```
zspencer: I found this useful for debugging purposes, but I think I should probably revert this or mark them as private explicitly?

lib/message.rb
```diff
@@ -71,8 +71,31 @@ def as_json(options = {})
...
+    def events?
+      !events.nil? && !events.empty?
+    end
+
+    def events=(events)
+      unless events.respond_to?(:map)
+        raise TypeError, "unable to iterate over #{events}, events must respond to #map"
+      end
+
+      @events = events.map do |event_data|
```
khamidou: Ah, I should have mentioned this – the event field is read-only so we don't need to be able to assign objects fields to it.

khamidou:
Zee,

Thanks for the PR. I just realized reading it I didn't mention that the events field on a message is basically read-only (it exposes the events attached to a message). As such, we don't need to be able to assign individual events to it. Apart from this the PR LGTM.

lib/message.rb
zspencer: Ah! good to know :)

lib/restful_model_collection.rb
khamidou: Curious, could you expand on how this is useful for debugging?

zspencer: Sure! I do a lot of REPL driven exploration of the code, and in a few different cases as I was exploring the API I wanted to inject a different version of `_api` and/or `model_class` to get better insights into what was going on. Having these available allowed me to replace them without reaching in and changing the instance variable.

zspencer:
@khamidou - OK this is ready for re-review, repositioned the interface to emphasize that events are read-only.

khamidou:
LGTM, but should we add a short example in the README to show the new APIs?

zspencer:
Added example to the messages section of the README. I would like to reinvest the README documentation into an interactive ruby file that people can use to get familiar with the API instead of a document. 

Thoughts on that in a later ticket?

Issue #116, opened 2017-11-09
Include all supported fields for the Threads API


zspencer:
The threads API is currently missing a few fields: version, last_message_sent_timestamp, and last_message_received_timestamp. Make sure the parameters for Thread match the [Thread API docs](https://docs.nylas.com/reference#threads)

PR #136, opened 2017-11-14
Add missing fields (version, last_message_sent, last_message_received) to threads

zspencer:
Resolves https://github.com/nylas/nylas-ruby/issues/116

### DIFF
--- a/lib/nylas.rb
+++ b/lib/nylas.rb
@@ -9,7 +9,7 @@ require 'active_support/core_ext/hash'
 
 require 'account'
 require 'api_account'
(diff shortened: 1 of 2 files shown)


lib/nylas.rb
```diff
@@ -9,7 +9,7 @@
 
 require 'account'
 require 'api_account'
-require 'api_thread'
```
khamidou: Ah, nice! No more naming conflicts!

khamidou:
LGTM, thanks for fixing!

Issue #115, opened 2017-11-09
Include all supported fields for the Account API


zspencer:
Currently, the Account model doesn't support the Name, email address, provider, organization unit, or billing state parameters, all of which are described in the [Accounts API](https://docs.nylas.com/reference#accounts)

zspencer:
I think this is unnecessary. Turns out there's two different representations of an account in Nylas, and those are split between two models:

`APIAccount` is a representation that comes from `GET /account` or calling `GET /a/:app_id/accounts/:account_id` when using a self-hosted version of the API and it *does* have all the fields described in the API documentation.

`Account` appears to be a near-but-not-quite-subset of the fields described in the docs for `GET /account` and is missing the `email` parameter exposed via the API.

PR #137, opened 2017-11-14
Add missing fields to Account model

zspencer:
Resolves https://github.com/nylas/nylas-ruby/issues/115


khamidou:
LGTM, thanks for fixing this!

PR #138, opened 2017-11-20
Update lockfiles after changing the version

zspencer:
When we upgrade the version, our cross-compatible gemfiles will need to have their lockfile update as well.

Otherwise CI will get nervous because we're trying to load a new version of the gem that isn't specified in the lockfile.


khamidou:
Whoops 😅

PR #143, opened 2017-11-21
Remove surprising behavior of not requiring enter on user input

zspencer:
When I was running the feature tests, I got some surprising behavior where it would move forward two steps instead of one.

This is because it wasn't waiting for me to hit the return key to validate my input, so when I hit "Y" + Enter it moved twice instead of once.

I've changed to use `gets` instead, so that future contributors don't run into the same behavior.


khamidou:
LGTM, thanks for fixing this!

PR #144, opened 2017-11-22
Don't require em-http and yajl if streaming is not needed.

marshall-lee:
`Nylas::API#delta_stream` is not always needed but it requires to install several gems.
This is an attempt to fix this.

### DIFF
--- a/gemfiles/Gemfile.rails4.lock
+++ b/gemfiles/Gemfile.rails4.lock
@@ -2,9 +2,7 @@ PATH
   remote: ..
   specs:
     nylas (3.2.0)
-      em-http-request (~> 1.1, >= 1.1.3)
       rest-client (>= 1.6)
-      yajl-ruby (~> 1.2, >= 1.2.1)
 
 GEM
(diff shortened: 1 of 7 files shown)


zspencer:
@khamidou what are your thoughts on this?

marshall-lee:
@zspencer are you going to release this change as 3.x or as 4.0?

zspencer:
@marshall-lee well, your branch probably could merge cleanly into the 3.X-master branch, if you make a PR against that branch I'd happily merge it so it comes out as part of 3.3 (if we wind up releasing it)

zspencer:
@marshall-lee - That said, I think we will want to maintain backwards compatibility with 3.X, so I'd prefer if we offered a `nylas-without-datastreams` gem that people who don't want event machine and yajl can opt into that without breaking upstream consumers. 

Given that, I think it's pretty unlikely we'll be formally releasing this as part of the 3.X major release, but  if you run off the 3.X branch in your gemfile it should work?

PR #147, opened 2017-11-23
LEARNING SPIKE DO NOT MERGE - Begin fleshing out a V2 API Client for the Contacts API

zspencer:
This is an initial stab at a way to make the SDK support a V2 API while leaving everything backwards compatible with the V1::API, which may or may not change by the time we get ready to release 4.0.0

Needs a bit more work to flesh out the tests and documentation and implementation, but this feels like a reasonable direction to me at this time.


lib/nylas.rb
```diff
@@ -21,8 +21,20 @@
...
+require 'nylas/v2'
 
 module Nylas
+  # @param version [String] Indicates which version of the API you want to use.
+  # @param app_id [String] The application's OAuth Client ID. Can be found on
+  #                        {https://dashboard.nylas.com/applications/ your applications Nylas API Dashboard}
+  # @param app_secret [String] The applications OAuth Client Secret. Can be found on
+  #                            {https://dashboard.nylas.com/applications/ your applications Nylas API
+  #                            Dashboard}
+  # @return [Nylas::V1::SDK, Nylas::V2::SDK]
```
zspencer: Missing the auth_token param, need to add it here.

lib/nylas/v2/model.rb
```diff
@@ -0,0 +1,174 @@
+module Nylas
+  module V2
+    class Registry
```
zspencer: Could probably use some unit tests and a usability pass here.

lib/nylas/v2/model.rb
```diff
@@ -0,0 +1,174 @@
...
+
+    module Model
+      def self.included(model)
+        model.extend(ClassMethods)
+      end
+
+      def initialize(**initial_data)
+        initial_data.each do |attribute_name, value|
+          self.send(:"#{attribute_name}=", value)
+        end
```
zspencer: I think we can probably do `self.attributes[name] = value` now that we have an `Attributes` class and we won't have to use send anymore.

lib/nylas/v2/model.rb
```diff
@@ -0,0 +1,174 @@
...
+
+        def define_accessors(name)
+          define_method :"#{name}" do
+            attributes[name]
+          end
+
+          define_method :"#{name}=" do |value|
+            attributes[name] = value
+          end
```
zspencer: We'll want to make sure we make YARD recognize that these methods are defined when attribute and has_n_of_attribute are called.

zspencer:
Done using this for reference.

PR #150, opened 2017-11-28
Minor syntactical cleanup to migrate to modern ruby hash syntax.

marshall-lee:
I wanted to force-push but #149 has already been merged :)
Ref #149

### DIFF
--- a/lib/nylas/restful_model.rb
+++ b/lib/nylas/restful_model.rb
@@ -61,8 +61,6 @@ module Nylas
     end
 
     def update(http_method, action, data = {}, params = {})
-      http_method = http_method.downcase
-
       @_api.execute(
         http_method,
         url(action),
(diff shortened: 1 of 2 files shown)


lib/nylas/restful_model.rb
```diff
@@ -61,8 +61,6 @@ def as_json(options = {})
     end
 
     def update(http_method, action, data = {}, params = {})
-      http_method = http_method.downcase
```
zspencer: Hmm, we sure we don't need this?

marshall-lee: @zspencer before the recent changes it was needed to pass a correct method name to `send`. now we pass `http_method` as a `RestClient` parameter so it does all the stuff with downcasing.

zspencer: 👍 awesome! That makes sense to me!

PR #151, opened 2017-11-28
Split out a `nylas-streaming` gem

zspencer:
This lays the foundation for a future `nylas-rails` gem while keeping
our build process and core dependencies lightweight.

We can now have `nylas` be the rest/http client while `nylas-streaming`
includes the event machine and yajl functionality so that users who
aren't using those don't pull in unnecessary dependencies or run into
dependency resolution conflicts.

hat tip to @marshall-lee for the inspiration for this.


adarsh:
I like this plan 👍

PR #152, opened 2017-11-29
Remove implicit dependency on ActiveSupport

zspencer:
For some reason, we had a require for `active_support` in lib/nylas,
which meant that even though we didn't declare a dependency on
ActiveSupport, we assumed that you would be using it.

This removes that in favor of a small class that casts hashes to query
strings.

We may be able to get rid of this if we leverage RestClient for
URL building.

### DIFF
--- a/lib/nylas.rb
+++ b/lib/nylas.rb
@@ -2,7 +2,7 @@ require 'json'
 require 'rest-client'
 
 require 'ostruct'
-require 'active_support/core_ext/hash'
+require_relative 'nylas/to_query'
 
 require 'nylas/account'
 require 'nylas/api_account'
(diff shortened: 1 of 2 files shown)


khamidou:
LGTM, thanks for fixing this! One question though – what's bad about depending on ActiveSupport? Won't most Ruby projects use it anyway?

zspencer:
There's nothing really bad about ActiveSupport, in fact I really like it. The problem comes when you aren't declaring which version you are dependent on, or if you *do* declare a dependency on it, and then the next version removes the feature(s) you were relying on it can cause upgrade difficulties for your consumers.

From my perspective, the Active<X> series of gems are intended for applications, and libraries should avoid using them unless they have a strong need as they do churn features between major versions and you don't want to be in a spot where you have to do a bunch of work to support Rails 6.

PR #158, opened 2017-12-08
Model v1 of the contact

zspencer:
This also allows us to flag models as read only and not searchable and inform the user if they try to use a feature that doesn't work


spec/v2/email_address.rb
```diff
@@ -1,25 +0,0 @@
-module Nylas
```
zspencer: This was a dupe I didn't notice til now.

lib/nylas/v2.rb
```diff
@@ -1,6 +0,0 @@
-require_relative 'v2/contact'
-
-module Nylas
-  module V2
-  end
-end
```
zspencer: Don't need this anymore since the only stuff that will be in v1 and v2 are the only models that differ, AKA the Contact model.

lib/nylas/v1/contact.rb
```diff
@@ -0,0 +1,20 @@
+module Nylas
+  module V1
```
khamidou: Hmm, I'm not sure I understand why we need this. The new contact model changes will only be in v4.x of the SDK, where we assume that everybody will be using the new version of the API (server-side). Could you explain why we'd need to have references to v1 and v2 of the API in the SDK?

zspencer: Nah we can kill it. I was mostly using it to grapple with a model that is read only and not searchable since some of the other stuff is. I was still a bit fuzzy on whether we wanted to force people to upgrade both the SDK and their API version, but I think it's cool.

zspencer:
I am going to do a "rubocop -a" pass after I merge this, which will fix 88% of the linting errors in one fell swoop; so I'm not going to kep trying to fix each and every error at this point.

PR #162, opened 2017-12-12
Accounts can be listed, activated, and deactivated

zspencer:
Adds Account Management API features  to 4.0


lib/nylas/types.rb
```diff
@@ -79,5 +79,14 @@ def cast(value)
       end
     end
     Types.registry[:string] = StringType.new
+
+    # Type for attributes represented as booleans.
+    class BooleanType < ValueType
+      # @param value [Object] Strictly casts the passed in value to a boolean (must be true, not "" or 1)
+      def cast(value)
```
zspencer: I am wondering if this is nil, it should be nil?

That way if we get a nil value in something that could be boolean, we don't assume it's false. Thoughts @khamidou?

lib/nylas/account.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    attribute :id, :string
+    attribute :account_id, :string
+    attribute :billing_state, :string
+    attribute :sync_state, :string
+
+    attribute :email, :string
+    attribute :trial, :boolean
+
+    def activate
```
khamidou: Hmm, I think we should keep the same terminology as our existing API – why not rename these methods `upgrade!` and `downgrade!`?

lib/nylas/http_client.rb
```diff
@@ -111,11 +117,11 @@ def default_headers
     rescue JSON::ParserError
       response
     end
+    inform_on :parse_response, level: :debug, also_log: { result: true }
```
khamidou: Curious, what does `inform_on` do?

lib/nylas/model.rb
```diff
@@ -18,22 +18,27 @@ def self.included(model)
     def save
       raise_if_read_only
       result = if id
-                 api.execute(method: :put, payload: attributes.serialize, path: resource_path)
+                 exec
```
khamidou: Hmm, did you mean to leave `exec` here?

lib/nylas/types.rb
khamidou: Hmm, I don't think of a case where this would happen. Maybe we should raise an exception?

zspencer: I can see it in more generic use cases, but not with your API as your API doesn't send non-true-values. Exception makes sense!

lib/nylas/http_client.rb
zspencer: inform_on logs the entire output of the `parse_response` if you have your NYLAS_LOG_LEVEL set to `debug` or greater verbosity. 

Very useful for diagnosis and exploration.

lib/nylas/model.rb
zspencer: ah, good catch! It's a no-op since it doesn't do anything, but leaving a system call lying out is not a good thing!

PR #165, opened 2017-12-16
Support Messages CRUD

zspencer:
Have some work to do with search and filtering still, but the entire message can be deserialized, including it's Event, File, Folder, and Labels objects.


khamidou:
LGTM!

PR #166, opened 2017-12-19
Expose Folder and Label CRUD actions

zspencer:


lib/nylas/api.rb
```diff
@@ -40,6 +40,16 @@ def accounts
       @accounts ||= Collection.new(model: Account, api: as(client.app_secret))
     end
 
+    # @return [Collection<Folder>] A queryable collection of Accounts
+    def folders
+      @accounts ||= Collection.new(model: Folder, api: self)
```
khamidou: Hmm, not sure I understand the `@accounts ||=` part.

spec/fixtures/opensource_accounts_endpoint.txt
```diff
@@ -1,11 +0,0 @@
-[
```
khamidou: Did you mean to delete all these fixtures?

lib/nylas/api.rb
zspencer: Copy-paste error. Good catch!

spec/fixtures/opensource_accounts_endpoint.txt
zspencer: Yeah, we aren't using them anymore. Instead, we're putting the fixtures inline in the tests so it's more clear what the inputs/outputs are.

PR #179, opened 2018-01-05
Contacts SDK returns picture

zspencer:


khamidou:
LGTM but I have a silly question – what if people just want to store the URL and not download it?

zspencer:
That's available at `contact.picture_url`

PR #180, opened 2018-01-22
Rename email_addresses field on contact to emails

zspencer:


khamidou:
LGTM, thanks for fixing this!

PR #181, opened 2018-01-30
Represent contacts birthday as a date

zspencer:


HayleyCAnderson:
Closing since this is well over a year old and it's unclear whether or why this change was desired. It can always be reopened with further explanation if necessary.

PR #186, opened 2018-06-25
Adding fields to the API does not break model initialization

zspencer:


lib/nylas/model/attributable.rb
```diff
@@ -21,6 +25,10 @@ def to_h
         attributes.to_h
       end
 
+      private def attribute_definitions
+        self.class.attribute_definitions
```
HayleyCAnderson: This is a little bit scary to me... I think I'd rather see the call to `class` where it's used rather than having to look in an instance method.

lib/nylas/model/attributable.rb
```diff
@@ -8,7 +8,11 @@ def self.included(model)
 
       def initialize(**initial_data)
         initial_data.each do |attribute_name, value|
-          send(:"#{attribute_name}=", value)
+          if attribute_definitions.key?(attribute_name)
```
HayleyCAnderson: I'm wondering if rather than doing this check, would it make sense to make use of `method_missing`? I think that could potentially be more clear about the intentionality and be more resilient.

lib/nylas/model/attributable.rb
34ndju: +1, don't think this is necessary either

spec/nylas/model_spec.rb
```diff
@@ -40,6 +40,10 @@ def example_instance_hash
       expect(instance.api).to eql(api)
     end
 
+    it "instantiates gracefully even if the api responds with additional fields" do
```
HayleyCAnderson: Is this the only place tests should be updated for this change?

lib/nylas/model/attributable.rb
zspencer: in my experience, method missing is a bit harder to trace down what is going on. This is checking that an attribute is defined, and if so, calling the method. I could see the argument made that this violates "tell don't ask" in that we want to tell the model to assign the value, but the level of indirection that method missing/metaprogramming introduces doesn't feel like a valuable trade off to me.

spec/nylas/model_spec.rb
zspencer: I'm not sure which places would be good to add this? This is testing it as close to the base implementation as possible, which means it's less likely to fail for weird reasons.

lib/nylas/model/attributable.rb
zspencer: Sounds reasonable. Gonna push this up.

lib/nylas/model/attributable.rb
HayleyCAnderson: Yeah, it's a tradeoff either way... As long as this is definitely the only place that's calling these attributes in this way, then that's fine.

zspencer: In cases where someone calls `contact.field_that_doesnt_exist="foo"` it will continue to raise an `undefined method` error; which feels like it's in line with the principle of least surprise.

spec/nylas/model_spec.rb
zspencer: Well, the other place this *could* bite us is when someone using `update` on a model; which is only called by users of the library. I think in those cases it's probably good to raise an error, as we want to make sure people aren't trying to update attributes that don't exist. That said a more intent-revealing error could be more appropriate.

zspencer: @HayleyCAnderson - I've changed update to raise an exception if a user attempts to update a field that is not part of the API schema supported by the gem.

PR #187, opened 2018-07-22
Finding a model works when new fields are added to the API

### LINKED ISSUE #188: "api is not defined as an attribute on" warnings on enumeration
I'm seeing errors like this when using the `.each` iterator on any `Collection` object:

```
W, [2018-07-27T13:56:02.919881 #31343]  WARN -- : api is not defined as an attribute on Nylas::Message
```
And just to confirm that I'm not totally crazy, it looks like you are too: https://travis-ci.org/nylas/nylas-ruby/jobs/406871743

The source of the issue is merging in the `api` instance to the model initialize method: 

https://github.com/nylas/nylas-ruby/blob/b8623d4721abce785451372ff69b52e78533c88e/lib/nylas/collection.rb#L65

During the model initialize method in the `Attributable` decorator, it checks for that field as an attribute definition:

https://github.com/nylas/nylas-ruby/blob/b8623d4721abce785451372ff69b52e78533c88e/lib/nylas/model/attributable.rb#L9-L17

While that's an `attr_accessor` on `Model`, it's not an `attribute` class definition and fails that check.

I'm not sure what the best course of action would be here (hence the issue instead of PR). You could whitelist that attribute, define it on the `Model` as a default for all models, or not merge it and pass it in some other way. I'm not sure what's best, but if you have some opinion about it, I can put together a PR.

zspencer:
When finding a particular record after the API had new fields added to
it, exceptions would be raised as the model would reload itself and
attempt to set a field that doesn't exist.

By making reload use the same logic as instantiation, we're able to
avoid exceptions being raised on find when new fields are added to the
API

This covers an additional use case from https://github.com/nylas/nylas-ruby/pull/186, which slipped past me.

This also resolves #188

### DIFF
--- a/lib/nylas/model/attributable.rb
+++ b/lib/nylas/model/attributable.rb
@@ -7,12 +7,12 @@ module Nylas
       end
 
       def initialize(**initial_data)
-        merge(**initial_data)
+        assign(**initial_data)
       end
 
-      protected def merge(**data)
+      protected def assign(**data)
         data.each do |attribute_name, value|
-          if self.class.attribute_definitions.key?(attribute_name)
+          if respond_to?(:"#{attribute_name}=")
             send(:"#{attribute_name}=", value)
           else
             Logging.logger.warn("#{attribute_name} is not defined as an attribute on #{self.class.name}")
--- a/lib/nylas/model/attributable.rb
+++ b/lib/nylas/model/attributable.rb
@@ -7,7 +7,11 @@ module Nylas
       end
 
       def initialize(**initial_data)
-        initial_data.each do |attribute_name, value|
+        merge(**initial_data)
+      end
+
+      protected def merge(**data)
+        data.each do |attribute_name, value|
           if self.class.attribute_definitions.key?(attribute_name)
             send(:"#{attribute_name}=", value)
           else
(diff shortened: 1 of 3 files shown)


lib/nylas/model/attributable.rb
```diff
@@ -7,7 +7,11 @@ def self.included(model)
       end
 
       def initialize(**initial_data)
-        initial_data.each do |attribute_name, value|
+        merge(**initial_data)
+      end
+
+      protected def merge(**data)
```
HayleyCAnderson: This doesn't do what I would call merging afaict - would `assign_attributes` or something make more sense?

spec/nylas/model_spec.rb
```diff
@@ -42,6 +42,14 @@ def example_instance_hash
     end
   end
 
+  describe "#reload" do
+    it "does not explode if using an invalid attribute" do
+      allow(api).to receive(:execute).and_return(fake_attribute: "not real")
+      instance = FullModel.from_json('{ "id": 1234 }', api: api)
+      expect { instance.reload }.not_to raise_error
```
HayleyCAnderson: I think it would be nice for this to include assertions that `id` is set on the model but `fake_attribute` is not. I think it would also be nice to include a test where an attribute does get set successfully on reload.

zspencer: Yea, I can add more fields to the model to show that the other stuff gets set.

Issue #189, opened 2018-09-04
Support expanded view when using find on messages


pfista:
If I wanted to get a specific message using the expadned view to see the message id header, I'm unable to do so because find doesn't support constraints or views:

```
puts client.messages.expanded.find(msg[:id]).to_h[:headers][:message_id]
```

`headers` here will always be nil

dem:
@pfista as per current code of `Model` there is no way to fetch expanded view on any model instance, using `reload` or `find` methods. For temporarily solution we used monkey patching:
```
class Nylas::Message
  def expanded
    return self unless headers.nil?

    assign(api.execute(method: :get, path: resource_path, query: { view: 'expanded' }))
    self
  end
end
```

And it become possible to fetch single expanded message in way it was in 3.x version:
`puts client.messages.find(msg[:id]).expanded.headers.message_id`

It is not optimal because of 2 calls to nylas api, but should work.

bmulholland:
This just caught me. Would be really nice if we could do `nylas.messages.expanded.find` just like `nylas.meessages.expanded.where`. More documentation on using expanded views in Ruby would be great, too.

PR #192, opened 2018-10-12
Add expanded method to message

dem:
When message fetched using `find`, it is always not expanded. Allow to use `expanded` method of message to return expanded version always.

If headers nil, just re-fetch expanded view of message.

It allows to run following:
```
message.expanded.headers.in_reply_to
```

Fix issue https://github.com/nylas/nylas-ruby/issues/189

### DIFF
--- a/lib/nylas/message.rb
+++ b/lib/nylas/message.rb
@@ -41,5 +41,12 @@ module Nylas
     def unread?
       unread
     end
+
+    def expanded
+      return self unless headers.nil?
+
+      assign(api.execute(method: :get, path: resource_path, query: { view: "expanded" }))
+      self
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


HayleyCAnderson:
I wonder if we could in the future redesign this to be a bit more reusable or efficient for customers, but this seems like a great fix to known issue for customers! Thanks so much for contributing!

dem:
@HayleyCAnderson I think it is good to make also following code to work
```
message = api.messages.expanded.find(id)
```
where `message` will be expanded in the end. But this requires a more complex work to do.

Together with `message.expanded` both options will work great. We use `message.expanded` to re-fetch on message only in part of processed messages, where we need this data. And also it add compatibility with 3.2, where it works.

PR #193, opened 2018-10-14
Fix encoding issue when download attachments

dem:
When downloading file exception raised:
`Encoding::UndefinedConversionError: "\x84" from ASCII-8BIT to UTF-8`

It fix following code to run:
```
message = api.messages.find(id)
content = message.files[0].download
```

PR #194, opened 2018-10-14
Transfer api attribute to included objects

dem:
When some model object initialized, also initialized and included dependent model objects.
For example when **message** is initialized, also will be initialized:  **events**, **files**, **folder** and **labels**. But this objects api will be nil, and it is not possible and easy to start using execute api methods.

Example:
```
content = message.files[0].download
```

Download will fail, api is not initialized for file.

This PR add `Transferable` class extension, using it it is possible to specify transfer of attributes to dependent objects using following notation:
```
module Nylas
  class Message
...
  end
end
```

Transfer api added to message, draft, thread. It is rspec covered.

### DIFF
--- a/lib/nylas/draft.rb
+++ b/lib/nylas/draft.rb
@@ -31,6 +31,8 @@ module Nylas
     attribute :folder, :label
     has_n_of_attribute :labels, :label
 
+    transfer :api, to: %i[events files folder labels]
+
     def send!
       save
       execute(method: :post, path: "/send", payload: JSON.dump(draft_id: id, version: version))
--- a/lib/nylas/message.rb
+++ b/lib/nylas/message.rb
@@ -34,6 +34,8 @@ module Nylas
     attribute :folder, :label
     has_n_of_attribute :labels, :label
 
+    transfer :api, to: %i[events files folder labels]
+
     def starred?
       starred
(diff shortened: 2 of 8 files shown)


dem:
@pswaminathan  @HayleyCAnderson Any updates on this pull request or alternative fix for issue? And for https://github.com/nylas/nylas-ruby/pull/193?

Currently we waiting for solutions in `gem 'nylas'` to fix problems with api attribute on included objects and fix for download method. And it is stops us to move from 3.x to 4.x `gem 'nylas'` . 

We could do monkey patch, but it looks like change used, but working car to new car with broken brakes and fix it with adhesive tape on-the-go :-)

pswaminathan:
My apologies for the delay here! Got a bit busy and dropped the ball on this. Let me take a look.

jpcamara:
Any update on this @pswaminathan ? I'm hitting the same issue as @dem

dem:
@jpcamara  Download method should be fixed in this pull request https://github.com/nylas/nylas-ruby/pull/193
If you have chance, please, test it.

We are using v4, but with 3-4 monkey patches.

HayleyCAnderson:
Looks great, thanks so much for the contribution!

PR #195, opened 2018-10-14
Fix wrong type for folder attribute in message, draft

dem:
It looks like mistype, but i think, for folder attribute type should be `folder`
PR fix it for message and draft models

### DIFF
--- a/lib/nylas/draft.rb
+++ b/lib/nylas/draft.rb
@@ -28,7 +28,7 @@ module Nylas
 
     has_n_of_attribute :events, :event
     has_n_of_attribute :files, :file
-    attribute :folder, :label
+    attribute :folder, :folder
     has_n_of_attribute :labels, :label
 
     def send!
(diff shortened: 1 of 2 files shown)


HayleyCAnderson:
Strange! I'm surprised this didn't get caught by testing. Were you able to confirm the bug some other way?

dem:
@HayleyCAnderson I think it did not caught by tests because `folder` and `label` types are mostly the same in attributes names/types. Look into:
https://github.com/nylas/nylas-ruby/blob/master/lib/nylas/label.rb
https://github.com/nylas/nylas-ruby/blob/master/lib/nylas/folder.rb

But if you try to reload from api, code will fail i suppose (did not try it):
```
folder = message.folder
folder.reload
```
It will goes to the wrong endpoint

HayleyCAnderson:
Unfortunate that it's not covered more. Thanks for the contribution!

PR #205, opened 2019-02-22
add account revoke-all method

snop-snov:
Implement revoke-all api method from Account Management spec:
https://docs.nylas.com/reference#revoke-all


singingwolfboy:
Could you rebase this pull request? That should make the tests pass.

Also, please include an optional argument for the `keep_access_token` param available in the API.

snop-snov:
fixed

PR #206, opened 2019-02-25
Add scopes arg for native authentication

singingwolfboy:
Add `scopes` argument to `Nylas::API#authenticate` for [selective sync](https://docs.nylas.com/docs/how-to-use-selective-sync).

Python SDK: https://github.com/nylas/nylas-python/pull/106

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ api.authenticate(
 )
 ```
 
+If you do not pass any scopes, then the SDK will default to using all of them.
+
 ### Handling Errors
 The Nylas API uses conventional HTTP response codes to indicate success or failure of an API request. The ruby gem raises these as native exceptions.
 
--- a/README.md
+++ b/README.md
@@ -41,6 +41,24 @@ We support Rails 4.2 and above. A more detailed compatibility list can be found 
 
 Examples are located in the [examples](./examples) directory. Examples in plain ruby are in [examples/plain-ruby/](./examples/plain-ruby). They are grouped by the API endpoints they interact with.
 
+## Scopes
+
+The Nylas API allows you to set various scopes during the authentication process
+in order to use the [Selective Sync](https://docs.nylas.com/docs/how-to-use-selective-sync)
+feature. Currently, these scopes are `email`, `calendar`, and `contacts`.
+You can pass an array of scopes to `Nylas::API#authenticate`, like this:
+
+```ruby
+api = Nylas::API.new()
+api.authenticate(
+    name: 'fake',
+    email_address: 'fake@example.com',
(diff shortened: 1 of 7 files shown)


lib/nylas/native_authentication.rb
```diff
@@ -7,16 +7,19 @@ def initialize(api:)
...
+      scopes = scopes.join(",") unless scopes.is_a?(String)
       code = retrieve_code(name: name, email_address: email_address, provider: provider,
-                           settings: settings, reauth_account_id: reauth_account_id)
+                           settings: settings, reauth_account_id: reauth_account_id, scopes: scopes)
 
       exchange_code_for_access_token(code)
     end
 
-    private def retrieve_code(name:, email_address:, provider:, settings:, reauth_account_id:)
+    private def retrieve_code(name:, email_address:, provider:, settings:, reauth_account_id:, scopes: nil)
```
HayleyCAnderson: Shouldn't this not have a default?

spec/nylas/native_authentication_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      ).and_return(access_token: "fake-token")
+      api = Nylas::API.new(client: client)
+
+      expect(
+        api.authenticate(
+          name: 'fake',
+          email_address: 'fake@example.com',
+          provider: :gmail,
+          settings: {},
+          scopes: ["email"]
```
HayleyCAnderson: I think since multiple scopes and string scopes are allowed, there should be tests for 1. array of 2 scopes 2. string of scope(s)

lib/nylas/native_authentication.rb
singingwolfboy: Fixed! Thanks.

spec/nylas/native_authentication_spec.rb
singingwolfboy: I added more tests.

README.md
```diff
@@ -41,6 +41,24 @@ We support Rails 4.2 and above. A more detailed compatibility list can be found
 
 Examples are located in the [examples](./examples) directory. Examples in plain ruby are in [examples/plain-ruby/](./examples/plain-ruby). They are grouped by the API endpoints they interact with.
 
+## Scopes
+
+The Nylas API allows you to set various scopes during the authentication process
+in order to use the [Selective Sync](https://docs.nylas.com/docs/how-to-use-selective-sync)
+feature. Currently, these scopes are `email`, `calendar`, and `contacts`.
```
HayleyCAnderson: I think you should also specify that not passing scopes will default to all.

PR #207, opened 2019-02-26
Relax bundler version requirements

singingwolfboy:


HayleyCAnderson:
I ran into this with Netlify recently. It's a very frustrating problem, and you can't get away with a change like this because often dependencies have bundler version requirements. In this case, your change worked for 3 of the 4 Gemfile options but not the Rails 4 one because even the latest version of Rails 4 requires bundler < 2.0.

I think this solution is the most likely to work: https://docs.travis-ci.com/user/languages/ruby/#bundler-20 (namely the second example which says "If you are using Ruby 2.3.x but wish to stay on Bundler 1.x (e.g., for dependency reasons such as Rails 4.2.x)")

Some additional context and options here: https://travis-ci.community/t/bundle-is-not-installed-for-ruby-2-3/1641/7

singingwolfboy:
I found a better way to resolve this problem in #208, so I'm closing this pull request.

PR #213, opened 2019-03-18
adding ip address endpoint and test

amanda103:
didn't mean to include the version change/changelog.md

### DIFF
--- a/lib/nylas/account.rb
+++ b/lib/nylas/account.rb
@@ -33,11 +33,6 @@ module Nylas
       response[:success]
     end
 
-    def ip_addresses
-        response = execute(method :get, path "#{resource_path}/ip_addresses")
-        response[:success]
-    end
-
     def self.resources_path(api:)
       "/a/#{api.app_id}/accounts"
     end
--- a/lib/nylas/account.rb
+++ b/lib/nylas/account.rb
@@ -33,6 +33,11 @@ module Nylas
       response[:success]
     end
 
+    def ip_addresses
+        response = execute(method :get, path "#{resource_path}/ip_addresses")
+        response[:success]
+    end
+
     def self.resources_path(api:)
       "/a/#{api.app_id}/accounts"
     end
--- a/lib/nylas/api.rb
+++ b/lib/nylas/api.rb
@@ -103,7 +103,8 @@ module Nylas
     # @return [Hash]
     # hash has keys of :updated_at (unix timestamp) and :ip_addresses (array of strings)
     def ip_addresses
-        response = client.as(access_token).get(path: "/ip_addresses")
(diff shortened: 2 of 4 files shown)


lib/nylas/account.rb
```diff
@@ -33,6 +33,11 @@ def revoke_all(keep_access_token: nil)
       response[:success]
     end
 
+    def ip_addresses
+        response = execute(method :get, path "#{resource_path}/ip_addresses")
```
HayleyCAnderson: Looks like you're missing colons here

lib/nylas/account.rb
```diff
@@ -33,6 +33,11 @@ def revoke_all(keep_access_token: nil)
       response[:success]
     end
 
+    def ip_addresses
```
HayleyCAnderson: What are you expecting to be returned from this method?

lib/nylas/api.rb
```diff
@@ -99,6 +99,13 @@ def revoke(access_token)
       response.code == 200 && response.empty?
     end
 
+    # Returns list of IP addresses
+    # @return [Boolean]
```
HayleyCAnderson: What are you expecting to be returned here? This says both a boolean and a list of ip addresses

lib/nylas/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Nylas
-  VERSION = "4.2.4"
+  VERSION = "4.3.0"
```
HayleyCAnderson: Lmk if you need help removing these changes from this PR

lib/nylas/account.rb
amanda103: something that looks like this {"updated_at": timestamp, "ip_addresses": [1234, 3457, 2213]}

amanda103: what is the correct variable name for that?

HayleyCAnderson: Hm okay. Yeah so this is getting into the problem with how this gem was written. I think you're going to have to make more, much less obvious changes including making a new model. We might have to pair on it.

lib/nylas/api.rb
HayleyCAnderson: If you look above at this `@labels ||= Collection.new(model: Label, api: self)` type situation a few lines above, I think that's basically what you're going to want to do here. So you'll have to start with making an `IPAddress` model. Maybe see if you feel comfortable trying to follow that path and then lmk if you want to pair.

amanda103: Sounds good - I'll work on this tomorrow and ping you if I get stuck.

amanda103:
I think I was confused on what tests looked like in ruby. Can you point me to where those tests would go for this endpoint?

PR #214, opened 2019-03-26
adding ip_addresses endpoint

amanda103:


lib/nylas/api.rb
```diff
@@ -99,6 +99,15 @@ def revoke(access_token)
       response.code == 200 && response.empty?
     end
 
+    # Returns list of IP addresses
+    # @return [Hash]
+    # hash has keys of :updated_at (unix timestamp) and :ip_addresses (array of strings)
+    def ip_addresses
+        path = "/a/"+app_id+"/ip_addresses"
```
HayleyCAnderson: "string/#{ var }/string" to avoid the + situation

amanda103: Done!

Issue #199, opened 2018-11-27
For events, the 'when' attributes works only for events in "timespan" mode


frederickdubois:
The `when`  attribute will have different return formats from the API depending on the type of date/time used for the event... Only the "timespan" works currently via the nylas-ruby library...

Examples values for different possible `when` attributes:

```ruby
# when 'timespan', I'm ok!
{:object=>"timespan", :start_time=>1543456800, :end_time=>1543458600}

# when 'date', I don't have the date...
{:object=>"date", :start_time=>0, :end_time=>0}

# when 'datespan', I don't have any of the 2 dates...
{:object=>"datespan", :start_time=>0, :end_time=>0}
```

Testing directly with the API we can see the different return formats... 
https://docs.nylas.com/reference#get-events

arunagw:
Closing this issue now as #215 solved it. @frederickdubois Please let us know if it's still an issue. Thanks

PR #215, opened 2019-04-01
Add Nylas::When for generic time object handling

trozamon:
This PR adds support for `when` in `Nylas::Event` for types other than `timespan`. I did this in the quickest/most obvious way to me - if there's a better way to implement this, I'll be happy to make any changes.

This is related to #199 and may fix that issue.

### DIFF
--- a/lib/nylas.rb
+++ b/lib/nylas.rb
@@ -35,6 +35,7 @@ require_relative "nylas/rsvp"
 require_relative "nylas/timespan"
 require_relative "nylas/web_page"
 require_relative "nylas/nylas_date"
+require_relative "nylas/when"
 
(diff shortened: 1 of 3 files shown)


HayleyCAnderson:
Thanks so much for the contribution! It looks great. Do you believe this is covered by existing tests or are there tests you could add?

trozamon:
Yeah, I think this is covered by existing tests. `message_spec.rb` tests `when` and makes sure it still behaves like a timespan.

HayleyCAnderson:
I think it would be great to add tests for the various other types that were added, but given the existing limited scope of the message testing, I won't consider that a blocker. Thanks for your contribution!

PR #219, opened 2019-04-12
adding source attribute to contact model

amanda103:
customer that is using our ruby sdk has requested this change

### DIFF
--- a/lib/nylas/contact.rb
+++ b/lib/nylas/contact.rb
@@ -30,6 +30,7 @@ module Nylas
     attribute :office_location, :string
     attribute :notes, :string
     attribute :web_page, :web_page
+    attribute :source, :string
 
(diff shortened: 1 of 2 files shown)


amanda103:
question: this will need a new release, yes? If so, which number do i bump?

HayleyCAnderson:
If they want to upgrade to it now, yeah. Same process as before. Can walk through it on slack if needed? If you mean which number in the version, should be middle number since you're adding a feature.

Looks like there actually is testing for this, so you should add to that: https://github.com/nylas/nylas-ruby/blob/c9687b93d57de4e1a334b0c9d268ce421fa04b09/spec/nylas/contact_spec.rb

amanda103:
@HayleyCAnderson I added it to the test! It passes now. Thanks for catching that :)

PR #225, opened 2019-06-27
Jnci content reorg ruby 062019

jamesneimanconsulting:
@pfista This is a continuation of the effort to modify the SDK github markdown. This also contains a link to a hidden topic in readme.io (https://docs.nylas.com/docs/ruby-quick-start), and the previous content was migrated to Ruby tabs in the hidden SDK examples topic in readme.io (https://docs.nylas.com/docs/sdk-examples) mainly to preserve them for future use, as discussed.


jamesneimanconsulting:
@pfista This PR has a new commit addressing the same issues raised with Node.js.

Issue #223, opened 2019-05-20
Support hosted authentication


pfista:
Add support for Hosted Authentication with a helper `url_for_authentication` method that returns a url that the user can be redirected to login through Nylas' Hosted Authentication flow.

kyletolle:
Hi @pfista. It was me and my team that asked for guidance on how to use Nylas Hosted Auth with v4 of the gem. The methods `url_for_authentication` and `token_for_code` are both used by our existing code which used v3.2.0, but these two methods don't seem to exist in v4. Thanks for the help in figuring out what changes we need to make to be able to use the Hosted Auth flow with v4.

arunagw:
Done in #292

PR #227, opened 2019-07-03
Support hosted authentication

phwebi:
This PR adds a helper method to generate urls for hosted authentication (requested in https://github.com/nylas/nylas-ruby/issues/223)

We need this so that we can support hosted authentication (https://docs.nylas.com/reference#oauthauthorize) using this `nylas-ruby` gem without having to make out own separate http requests.

### DIFF
--- a/lib/nylas/api.rb
+++ b/lib/nylas/api.rb
@@ -36,6 +36,21 @@ module Nylas
       )
     end
 
+    # @return [String] A url for hosted authentication
+    def hosted_authentication_url(redirect_uri, login_hint = "", options = {})
+      params = {
+        client_id: app_id,
+        redirect_uri: redirect_uri,
+        response_type: options.fetch(:response_type, "code"),
+        scope: options.fetch(:scope, "email"),
+        login_hint: login_hint
+      }
+
+      params[:state] = options[:state] if options.key?(:state)
+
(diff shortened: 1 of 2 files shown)


lib/nylas/api.rb
```diff
@@ -36,6 +36,22 @@ def authenticate(name:, email_address:, provider:, settings:, reauth_account_id:
...
     end
 
+    # @return [String] A url for hosted authentication
+    def hosted_authentication_url(redirect_uri:, login_hint: "", response_type: "code",
+                                  scope: "email", state: nil)
+      params = {
+        client_id: app_id,
+        redirect_uri: redirect_uri,
+        response_type: response_type,
+        scope: scope,
```
pk1m: should this be `scopes:` instead?

phwebi: I think it should be `scopes:`. Updated!

arunagw:
@phwebi Thanks for the PR. Can you please add a bit more in PR description to tell why we need this change and may be point to the documentation to show.

Thanks.

phwebi:
@arunagw Updated the description with a link to the hosted authentication docs.

joshkestenberg:
Any plans to get this merged? I just started using Nylas and this functionality would be ideal. It'd also be great to document this in the Nylas docs under Hosted Auth for Ruby SDK (which right now appears to describe native auth).

PR #228, opened 2019-07-09
implement exchange code for token

phwebi:
This PR implements the ability to exchange an authorization code for an oauth token.

When we make a request for hosted authentication, nylas provides us with a code that we need to exchange for an oauth token in order to successfully authenticate users. See https://docs.nylas.com/reference#oauthtoken.


lib/nylas/api.rb
```diff
@@ -36,6 +36,18 @@ def authenticate(name:, email_address:, provider:, settings:, reauth_account_id:
...
+    def exchange_code_for_token(code)
+      data = {
+        "client_id" => app_id,
+        "client_secret" => client.app_secret,
+        "grant_type" => "authorization_code",
+        "code" => code
+      }
+
+      response_json = execute(method: :post, path: "/oauth/token", payload: data)
+      response_json["access_token"]
```
pk1m: while testing, this was a symbol for me

arunagw:
@phwebi Thanks for the PR. Can you please update the PR description a bit more about what this change and why we need this change? May be point to documentation to show why we want this?

Thanks.

phwebi:
@arunagw Updated the description!

viral810:
Is this getting merged soon?

BenLloydPearson:
@phwebi, @viral810 Thank you for your patience on this. I've merged this PR and will create a minor gem release for this shortly.

PR #231, opened 2019-08-21
Allow setting `label_ids` on a message

arunagw:
Right now there is no way to set a `label` on a `message`.

API accepts `label_ids` as an array of string of label ids.

This PR also add `UPDATABLE_ATTRIBUTES` to just allow few attributes to
be updated allowed by API instead of having all attributes to be
updated.

### DIFF
--- a/lib/nylas/message.rb
+++ b/lib/nylas/message.rb
@@ -34,7 +34,10 @@ module Nylas
     has_n_of_attribute :events, :event
     has_n_of_attribute :files, :file
     attribute :folder, :folder
+    attribute :folder_id, :string
+
     has_n_of_attribute :labels, :label
+    has_n_of_attribute :label_ids, :string
 
     transfer :api, to: %i[events files folder labels]
 
@@ -44,6 +47,16 @@ module Nylas
 
     def unread?
       unread
+    end
+
+    UPDATABLE_ATTRIBUTES = %i[label_ids folder_id starred unread].freeze
+    def update(data)
+      unupdatable_attributes = data.keys.reject { |name| UPDATABLE_ATTRIBUTES.include?(name) }
+      unless unupdatable_attributes.empty?
+        raise ArgumentError, "Cannot update #{unupdatable_attributes} only " \
+                             "#{UPDATABLE_ATTRIBUTES} are updatable"
+      end
+      super(data)
     end
 
     def expanded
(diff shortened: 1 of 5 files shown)


lib/nylas/message.rb
```diff
@@ -34,7 +35,10 @@ class Message
     has_n_of_attribute :events, :event
     has_n_of_attribute :files, :file
     attribute :folder, :folder
+    attribute :folder_id, :string
+
     has_n_of_attribute :labels, :label
+    has_n_of_attribute :label_ids, :string
```
adarsh: Do we need to add `starred` and `unread` attributes here also? I'm not familiar with the DSL so I'm just looking at the pattern.

lib/nylas/message.rb
```diff
@@ -46,6 +50,15 @@ def unread?
       unread
     end
 
+    def update(data)
+      unupdatable_attributes = data.keys.reject { |name| UPDATABLE_ATTRIBUTES.include?(name) }
```
adarsh: What do you think about extracting a private method, something named like `def unupdatable_attributes?` to raise the exception. Then it would maybe look like

```ruby
def update(data)
  if unupdatable_attributes?(data)
    raise ArgumentError, # message
  else  
    super(data)
  end
end
```

?

lib/nylas/message.rb
arunagw: I am going to check with these attributes and see what's happening. If we don't need it I will remove them from here. Thanks for the point.

lib/nylas/message.rb
arunagw: Make sense.

lib/nylas/message.rb
arunagw: We do have `started` and `unread` define above the file so we are good. The `label_ids` was the missing one.

arunagw: They are actually here https://github.com/nylas/nylas-ruby/pull/231/files#diff-6027b1238ffa1b9852b7b40018edcbabL31-L32

lib/nylas/message.rb
arunagw: I have added a new class to filter attributes here https://github.com/nylas/nylas-ruby/pull/231/commits/e40e7cee254ced5f4b22d57b7932a65694fea700

I think we are going to need this class everywhere where we don't want to send unwanted params to server.

adarsh: Oh this is beautiful ✨

PR #232, opened 2019-08-23
Test against Rails 6 version

arunagw:
As Rails 6.0 is released we can should also add that to CI.
As Rails 6 only supports ruby 2.5 and above we need to 
add custom matrix for that in CI.

- Gemfile.rails6 added.
- >travis.yml update to run tests with Rails 6.

### DIFF
--- a/gemfiles/Gemfile.rails6
+++ b/gemfiles/Gemfile.rails6
@@ -1,0 +1,5 @@
+source 'https://rubygems.org'
+
+gem 'rails', '~> 6'
+
+gemspec path: "..", name: 'nylas-streaming'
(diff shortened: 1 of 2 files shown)


gemfiles/Gemfile.rails6
```diff
@@ -0,0 +1,5 @@
+source 'https://rubygems.org'
+
+gem 'rails', '~> 6'
+
+gemspec path: "..", name: 'nylas-streaming'
```
adarsh: I know why this is here but perhaps it might be a good idea to add a comment as to why this gemspec is included? What do you think?

Issue #183, opened 2018-03-11
Can't seem to set labels on a message


dv:
What's the best practice to update labels on a message? I can't seem to get it to work.

Trying to save a message:
```ruby
[75] pry(main)> message.save
Nylas::InvalidRequest: Invalid id: {u'display_name': u'All Mail', u'id': u'43jzq92uc9gqd8wreu4bw71ws', u'name': u'all'}
from /Users/david/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/nylas-4.2.0/lib/nylas/http_client.rb:141:in `handle_anticipated_failure_mode'
```

Trying to update the labels (`trash(api)` returns the `Nylas::Label` trash label):
```ruby
[76] pry(main)> message.update(labels: [trash(api)])
Nylas::InvalidRequest: Invalid id: {u'object': u'label', u'display_name': u'Trash', u'id': u'21hi7mcls49kn1yssqnw54vh4', u'account_id': xxxx'}
from /Users/david/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/nylas-4.2.0/lib/nylas/http_client.rb:141:in `handle_anticipated_failure_mode'
```

Just sending the IDs doesn't work either:
```ruby
[77] pry(main)> message.update(labels: [trash(api).id])
TypeError: labels Unable to cast 21hi7mcls49kn1yssqnw54vh4 to a Nylas::Label
from /Users/david/.rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/nylas-4.2.0/lib/nylas/model/attributes.rb:23:in `rescue in cast'
```

Considering that not changing anything and just calling `save` on a message already fails with an error, I'm leaning towards there being a bug in the ruby lib.

BenLloydPearson:
Hi @dv, #231 adds the ability to set labels on messages, you can now do this like so: `message.update(label_ids: ['label-1-id', 'label-2-id'])`

To your second point about having issues with .save, we have also noticed some inconsistencies with this function and we are currently looking into fixing it. I will let you know when we've fixed this.

BenLloydPearson:
Release 4.6.0 should fix the issues you're experiencing with .save. Please update to the latest version and let us know if you have any more issues with this.

PR #233, opened 2019-08-30
Fix an error occurring on `message.save`

arunagw:
Right now when we call `.save` on a message it raises an error 
because `labels` attributes also being sent to server along with 
other attributes. Server won't know anything about `labels` so it
was failing. This issues also mentioned here https://github.com/nylas/nylas-ruby/issues/183

This PR right now rejecting the `labels` attribute to be sent to server so 
we can save a `message`. 

In this PR also using a feature where we can define `exclude_when` call 
for certain actions.

### DIFF
--- a/lib/nylas/message.rb
+++ b/lib/nylas/message.rb
@@ -37,7 +37,7 @@ module Nylas
     attribute :folder, :folder
     attribute :folder_id, :string
 
-    has_n_of_attribute :labels, :label
+    has_n_of_attribute :labels, :label, exclude_when: [:saving]
     has_n_of_attribute :label_ids, :string
 
     transfer :api, to: %i[events files folder labels]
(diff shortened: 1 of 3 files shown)


lib/nylas/message.rb
```diff
@@ -37,7 +37,7 @@ class Message
     attribute :folder, :folder
     attribute :folder_id, :string
 
-    has_n_of_attribute :labels, :label
+    has_n_of_attribute :labels, :label, exclude_when: [:saving]
```
adarsh: This is neat, although I can't tell where it is defined.

arunagw:
I don't think we need any documentation changes for this as excluding this is an internal change.

We might remove this check when we implement change to allow `labels` to be set which gets converted into `label_ids` internally.

PR #234, opened 2019-08-30
Remove unused attribute `web_page` from contact

arunagw:
We are not using a single webpage anywhere so we can remove this from the attribute list


arunagw:
@pfista can you confirm this with the backend that we are not using this attribute anymore. The problem is when I am sending back this on `.save` it's giving me an error about unknown attribute.

Issue #200, opened 2018-12-05
No method for retrieving all contact groups


Che-Chris:
There is no method to get all `ContactGroup`s even though there is an API endpoint at `https://api.nylas.com/contacts/groups`.

PR #237, opened 2019-09-24
Add support for `/contacts/groups` endpoint

arunagw:
There is new endpoint for which we want to add support and it is `/contacts/groups`
This new endpoint is used to get all the `groups` from an account. Only list is allowed for the contact groups end point.

* Make `ContactGroup` class to be a model so we can fetch it standalone via api.
* Add spec to make sure we have all the attributes we want from the API.


arunagw:
This PR also requested here https://github.com/nylas/nylas-ruby/issues/200

lib/nylas/contact_group.rb
```diff
@@ -4,7 +4,15 @@ module Nylas
...
   class ContactGroup
-    include Model::Attributable
+    include Model
+    self.resources_path = "/contacts/groups"
+    self.creatable = false
+    self.listable = true
+    self.showable = false
+    self.filterable = false
+    self.updatable = false
+    self.destroyable = false
```
adarsh: What do you think about sorting this list either alphabetically or by which are `true` then `false`?

spec/nylas/contact_group_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  it "is not destroyable" do
+    expect(described_class).not_to be_destroyable
+  end
+
+  it "is listable" do
+    expect(described_class).to be_listable
+  end
+
+  it "is not showable" do
+    expect(described_class).not_to be_showable
```
adarsh: I ❤️ these matchers

lib/nylas/contact_group.rb
arunagw: taken care here 1da3b949e049ad20a30dbd7287624e3084328207

PR #239, opened 2019-09-30
Allow `message`, `thread` to be moved between folders

arunagw:


spec/nylas/message_spec.rb
```diff
@@ -186,6 +186,23 @@
     end
   end
 
+  describe "update_folder" do
+    it "move message to new `folder`" do
```
adarsh: Maybe `move` > `moves`?

spec/nylas/message_spec.rb
```diff
@@ -186,6 +186,23 @@
...
+      folder_id = "9999"
```
adarsh: Is this number Magic?

Would `rand(100)` also work? I'm not sure if this is a great idea, curious to hear what you think.

spec/nylas/message_spec.rb
```diff
@@ -186,6 +186,23 @@
...
+      api = instance_double(Nylas::API, execute: "{}")
+      message = described_class.from_json('{ "id": "message-1234" }', api: api)
+      allow(api).to receive(:execute)
+
+      message.update_folder(folder_id)
+
+      expect(api).to have_received(:execute).with(
+        method: :put,
+        path: "/messages/message-1234",
+        payload: { folder_id: "9999" }.to_json
```
adarsh: Here you could refer to the variable `folder_id` instead of the integer literal again - may make it less brittle.

spec/nylas/thread_spec.rb
```diff
@@ -63,6 +63,23 @@
...
+      api = instance_double(Nylas::API, execute: "{}")
+      thread = described_class.from_json('{ "id": "thread-1234" }', api: api)
+      allow(api).to receive(:execute)
+
+      thread.update_folder(folder_id)
+
+      expect(api).to have_received(:execute).with(
+        method: :put,
+        path: "/threads/thread-1234",
+        payload: { folder_id: "9999" }.to_json
```
adarsh: Same integer literal question here.

spec/nylas/message_spec.rb
arunagw: We don't need to use random here but we can use if we want as we are going to assert `folder_id` anyway in the expectation.

We have been using these magic numbers from a long time.

spec/nylas/message_spec.rb
arunagw: Yes. Done here 86e2a1fc6bb1a76b816c63819b05ee6cf707d325

spec/nylas/message_spec.rb
arunagw: Yes we can. Fixed here 86e2a1fc6bb1a76b816c63819b05ee6cf707d325

spec/nylas/thread_spec.rb
arunagw: Yes. Done here 86e2a1fc6bb1a76b816c63819b05ee6cf707d325

arunagw:
Thanks @adarsh for the review. I will merge this once CI is green.

PR #254, opened 2020-09-01
Add is_primary and other new Calendar attributes

monitron:
This PR adds support for new attributes that were recently added to the Nylas calendar API: `is_primary`, `location` and `timezone`. To match the convention established by the `read_only?` method, it also adds a `primary?` method which simply returns the value of `is_primary`.


spang (approved):
These look good to me. Thank you for your contribution @monitron !

PR #255, opened 2020-09-01
Handle new attributes added to API gracefully

monitron:
A couple of problems came up in our production app when Nylas added new attributes to API objects. One was a constant barrage of Rails logger warnings upon each request that drowned out other log traffic. A second more serious issue was that upon certain operations, the gem would raise exceptions about the unexpected attributes, causing failures in our code even though the requests were successful.

The [Nylas API docs](https://docs.nylas.com/reference#api-versioning) state:
> Nylas considers additive changes to be non-breaking. This means we may add additional fields to an API response object without releasing a new API version, so your application should handle this possibility.

I believe these changes to the gem are necessary in order to respect this very reasonable policy.

### DIFF
--- a/lib/nylas/model/attributable.rb
+++ b/lib/nylas/model/attributable.rb
@@ -27,8 +27,6 @@ module Nylas
         data.each do |attribute_name, value|
           if respond_to?(:"#{attribute_name}=")
             send(:"#{attribute_name}=", value)
-          else
-            Logging.logger.warn("#{attribute_name} is not defined as an attribute on #{self.class.name}")
           end
         end
       end
--- a/lib/nylas/model/attributes.rb
+++ b/lib/nylas/model/attributes.rb
@@ -17,6 +17,8 @@ module Nylas
 
       def []=(key, value)
         data[key] = cast(key, value)
+      rescue Nylas::Registry::MissingKeyError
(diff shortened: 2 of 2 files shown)


khamidou:
Hey @monitron thanks for sending the PR! It looks good to me but I had one question — what do you think about keeping the log message in? I think it'd be helpful for people to know that they need to upgrade (but also, at the same time they probably don't need to see this message hundreds of times).

Thank you!

monitron:
@khamidou thanks for reviewing! I removed the log message because it can be repeated thousands of times per hour, can start occurring at any time and we may not have a recourse to stop it from happening (as new attributes seem to start being included in API responses before a new version of the gem is released to handle them). As for upgrades, we have our dependency bot and emails from Nylas to let us know when changes are ready for us to adopt. So we'd definitely prefer to not have the log message if that's possible.

If that's not an option, it would be great to be able to turn it off, or to log it at a much lower importance than `warn`. Please let me know your thoughts.

spang:
Given that it's our policy that adding new attributes is not a breaking change, I think this is reasonable. Thank you for your contribution @monitron and for taking the time to contribute them back so that others can benefit!

monitron:
Thanks so much!!

PR #262, opened 2020-12-04
Remove folder on message.save for updates

nickbair-nylas:
## Issue
When updating a message in Outlook, we throw an error if the folder sub-object is present in the request.
```
/Users/nickbair/.rvm/gems/ruby-2.6.5/gems/nylas-4.6.2/lib/nylas/http_client.rb:166:in `handle_anticipated_failure_mode': Invalid id: {u'display_name': u'Junk Email', u'id': u'a8cquo7scqx8qpzd5hslmbeqz', u'name': u'spam'} (Nylas::InvalidRequest)
```

## The fix
If a folder is present when updating a message, store the folder's `id` in `folder_id` and remove `folder` from the request payload that is sent to the Nylas API.

### DIFF
--- a/lib/nylas/message.rb
+++ b/lib/nylas/message.rb
@@ -34,8 +34,8 @@ module Nylas
 
     has_n_of_attribute :events, :event
     has_n_of_attribute :files, :file
-    attribute :folder, :folder
     attribute :folder, :folder, exclude_when: [:saving]
+    attribute :folder_id, :string
 
     has_n_of_attribute :labels, :label, exclude_when: [:saving]
     has_n_of_attribute :label_ids, :string
--- a/lib/nylas/message.rb
+++ b/lib/nylas/message.rb
@@ -35,7 +35,7 @@ module Nylas
     has_n_of_attribute :events, :event
     has_n_of_attribute :files, :file
     attribute :folder, :folder
-    attribute :folder_id, :string
+    attribute :folder, :folder, exclude_when: [:saving]
 
     has_n_of_attribute :labels, :label, exclude_when: [:saving]
     has_n_of_attribute :label_ids, :string
--- a/lib/nylas/message.rb
+++ b/lib/nylas/message.rb
@@ -34,7 +34,7 @@ module Nylas
 
     has_n_of_attribute :events, :event
     has_n_of_attribute :files, :file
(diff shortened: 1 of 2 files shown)


khamidou:
Hey @nickbair-nylas — this is super impressive that you're already sending PRs on your first week!

The change looks good but I had a couple questions before approving it:
- what was the exact error we were seeing in prod?
- isn't updating the folder a valid use case for most providers?

Thanks! Happy to chat in person btw!

khamidou:
Looks good! Could we have a spec file to test the behavior new though? Thank you!

spec/nylas/message_spec.rb
```diff
@@ -174,7 +226,7 @@
         )
       )
     end
-
+
```
khamidou, line 229: Nitpick: you should remove this space

lib/nylas/message.rb
```diff
@@ -69,5 +69,26 @@ def expanded
...
+        method: :put,
+        payload: attributes.serialize(keys: allowed_keys_for_save),
+        path: resource_path
+      )
+    end
+
+    def handle_folder
+      return if folder.nil?
+      
+      if folder_id.nil? && !self.to_h.dig(:folder, :id).nil?
```
khamidou, line 86: Didn't know about `:dig`, that's a cool method!

lib/nylas/message.rb
```diff
@@ -69,5 +69,26 @@ def expanded
...
+
+      execute(
+        method: :put,
+        payload: attributes.serialize(keys: allowed_keys_for_save),
+        path: resource_path
+      )
+    end
+
+    def handle_folder
+      return if folder.nil?
```
khamidou, line 84: silly question but shouldn't we check in `self`?

nickbair-nylas: @khamidou Great question!  I thought is was required as well, but its not - https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/RedundantSelf.  For example, the linter is saying that `!self.to_h.dig(:folder, :id).nil?` does not need self.  I think having self is probably more readable, so happy to add that if you agree.

khamidou: Not a rubyist but if it works it works.

PR #264, opened 2020-12-17
Fix JSON parsing behavior

nickbair-nylas:
## Issue
In `http_client.rb` when parsing the response from an API call, if `JSON.parse` raises an error, the error is rescued and the raw response returned.  This causes an error downstream in `attributable.rb` and returns a non-specific error `wrong number of arguments (given 1, expected 0)`.

The one instance of a consistently reproducible parse error I was able to find using the Nylas API was an email with the unicode character `\ud83d`, thus is was used as the primary test case.  Technically speaking, `\ud83d` is not valid, but IMO, instead of raising an error, replacing the offending character or returning the string literal value is a better approach.

## Fix / Changes
1. Given an error when parsing JSON, no longer rescue the error.
2. Replace JSON with Yajl for parsing responses (also used in the Nylas streaming gem), instead of raising an error when parsing JSON that contains `\ud83d` it is replaced with a `?`.

### DIFF
--- a/lib/nylas/http_client.rb
+++ b/lib/nylas/http_client.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module Nylas
-  require 'yajl'
+  require "yajl"
 
   # Plain HTTP client that can be used to interact with the Nylas API sans any type casting.
   class HttpClient # rubocop:disable Metrics/ClassLength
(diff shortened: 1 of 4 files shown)


khamidou:
Looks great! Thank you for fixing this! Any idea what's up with the one failing build?

PR #269, opened 2021-01-11
Add message tracking to drafts.

nickbair-nylas:
Add support for message tracking to the drafts class.  Users who are not adding the tracking attribute to their draft will not hit the new code.


NHoppe (approved):
LGTM.

TravisCI is failing due to RuboCop rules.

PR #270, opened 2021-01-11
Add reply_to to NewMessage

nickbair-nylas:
The new message class (used for sending directly) in the SDK is missing the `reply_to` attribute (https://docs.nylas.com/reference#sending-directly).  Added this attribute and a test for the class.

### DIFF
--- a/spec/nylas/new_message_spec.rb
+++ b/spec/nylas/new_message_spec.rb
@@ -3,8 +3,6 @@
 require "spec_helper"
 
 describe Nylas::NewMessage do
-  let(:api) { FakeAPI.new }
-
   describe "#send!" do
     data = {
       reply_to_message_id: "mess-1234",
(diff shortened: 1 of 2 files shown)


viral810:
Thanks, @nickbair-nylas, this fixes the bug when users try to send emails directly and if they specify the `reply_to` field, its not sent as an empty field to API.

spec/nylas/new_message_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      cc: [{ email: "cc@example.com", name: "CC Example" }],
+      bcc: [{ email: "bcc@example.com", name: "BCC Example" }],
+      reply_to: [{ email: "reply-to@example.com", name: "Reply To Example" }],
+      subject: "A draft emails subject",
+      body: "<h1>A draft Email</h1>",
+      file_ids: [1234, 5678],
+      tracking: { opens: true }
+    }
+
+    it "sends the message directly" do
```
NHoppe, line 20: Not sure if I'm being paranoid here, but there's no test case for failure.

The [`send!` method](https://github.com/nylas/nylas-ruby/blob/cbb57d553bdfa82618400b3a814e5701dbb83fd9/lib/nylas/new_message.rb#L28-L30) calls `Api.execute` which is delegated to [`HttpClient.execute` method](https://github.com/nickbair-nylas/nylas-ruby/blob/dev-reply-to/lib/nylas/http_client.rb#L74-L89). However, the associated Spec file, [`http_client_spec.rb`](https://github.com/nickbair-nylas/nylas-ruby/blob/dev-reply-to/spec/nylas/http_client_spec.rb), isn't testing the failure handler used in [line 86 of `http_client.rb`](https://github.com/nickbair-nylas/nylas-ruby/blob/dev-reply-to/lib/nylas/http_client.rb#L86).

Shouldn't we ensure that `HttpClient` is handling the errors if they happen and, therefore, guarantee the reliability of the method `send!`?

nickbair-nylas: Hey @NHoppe, I think that's a valid concern.  One of my other PRs adds to the spec for `http_client` (https://github.com/nylas/nylas-ruby/pull/271/files), though the tests are specific to http status codes.  I can certainly add a test for the failure case of `send!` here if you think that's needed.  Thanks!

NHoppe: Hey @nickbair-nylas,

Thank you for understanding. 🙇 We should definitely check for failure, either in this PR or in another. Perhaps the recommended PR would be the [Add missing http status codes](https://github.com/nylas/nylas-ruby/pull/271/) due to context.

NHoppe (approved):
lgtm

PR #273, opened 2021-01-19
Truncate the returned filename so it's less than 256 characters

dominicj-nylas:
This ensures the filename will be compatible with rb_sysopen. See the [Clubhouse story](https://app.clubhouse.io/nylas/story/49502) for more details.

### DIFF
--- a/lib/nylas/file.rb
+++ b/lib/nylas/file.rb
@@ -56,7 +56,9 @@ module Nylas
     def retrieve_file
       response = api.get(path: "#{resource_path}/download")
       filename = response.headers.fetch(:content_disposition, "").gsub("attachment; filename=", "")
-      temp_file = Tempfile.new(filename, encoding: "ascii-8bit")
+      # The returned filename can be longer than 256 chars which isn't supported by rb_sysopen.
+      # 128 chars here is more than enough given that TempFile ensure the filename will be unique.
+      temp_file = Tempfile.new(filename[0..127], encoding: "ascii-8bit")
       temp_file.write(response.body)
       temp_file.seek(0)
       temp_file


bmulholland:
Won't this cut off the file extension (e.g. ".pdf")? I think it would be important to maintain that.

nickbair-nylas:
Hey @dominicj-nylas, Brendan is correct--if there is a file extension and the filename is sufficiently long, the file extension could be lost.  You could check for a file extension using regex, then split the string and trim the string to the left of the `.` in the filename.

Is the character limit specific to Ruby or is that specific to the user's OS?

Something like this:
```
reg = /(\.)(?!.*\.)/
if filename.match?(reg)
  arr = filename.split(reg)
  new_filename = "#{arr[0][0..127]}.#{arr.last}"
else
  new_filename = filename[0..127]
end
```

dominicj-nylas:
Appreciate the feedback @bmulholland and @nickbair-nylas!

`Tempfile` is already mangling the extension of the file when it makes sure it's unique i.e.

```
irb(main):003:0> temp_file = Tempfile.new("test.ext", encoding: "ascii-8bit")
=> #<Tempfile:/var/folders/4s/8b3t1qkd1hn42gpm3xwr44q80000gn/T/test.ext20210119-56019-1s31pcb>

irb(main):008:0> temp_file.path
=> "/var/folders/4s/8b3t1qkd1hn42gpm3xwr44q80000gn/T/test.ext20210119-56019-1s31pcb"
```

Take note that I'm not modifying the `filename` attribute but only the filename associated with the `Tempfile`.

Let me know what you think!

bmulholland:
Yeah that's a good point. I can see that cause a non-obvious bug in some of my code :\. There's a way to [ensure the extension is preserved ](https://makandracards.com/makandra/13719-ruby-how-to-ensure-a-tempfile-s-extension)- maybe that's a good idea?

I could create a followup PR if you'd like to keep this one clean.

dominicj-nylas:
@bmulholland I think your suggestion would introduce a new behavior in the SDK but if you think this could be helpful, I would go with a follow-up PR with rationale and perhaps some example code where you think that could be useful. The full filename including extension is still available via the `filename` attribute.

bmulholland:
Oh yeah, good point. Then actually my point about extensions isn't very important, and I was probably imagining bugs :). Thanks for taking the time to address my thoughts!

NHoppe (approved):
Considering @bmulholland concerns were the same as mine when I first saw the code, and it would actually change the SDK behaviour, I understand this code is good to be merged.

PR #276, opened 2021-02-12
Fix thread-safety issue in HTTP::CookieJar loading

bradrobertson:
HTTP::CookieJar uses unsafe class caching for dynamically loading cookie jar implementations. If 2 rest-client instances are instantiated at the same time, (for instance if you spawn 2 threads), then non-deterministic behaviour can occur whereby the Hash cookie jar isn't entirely loaded and cached. This _sometimes_ [raises an exception](https://github.com/sparklemotion/http-cookie/blob/master/lib/http/cookie_jar/abstract_store.rb#L21) because the Hash CookieJar can't be found (it's a race condition)

See:
- https://github.com/sparklemotion/http-cookie/issues/27
- https://github.com/sparklemotion/http-cookie/issues/6

Forcing an instantiation of the jar onload will force the Hash CookieJar to load before the system has a chance to spawn any threads.

Note this should technically be fixed in rest-client itself however that library appears to be stagnant so we're forced to fix it here. This object should get GC'd as it's not referenced by anything

### DIFF
--- a/lib/nylas.rb
+++ b/lib/nylas.rb
@@ -3,6 +3,21 @@
 require "json"
 require "rest-client"
 
+# BUGFIX
+#   See https://github.com/sparklemotion/http-cookie/issues/27
+#   and https://github.com/sparklemotion/http-cookie/issues/6
+#
+# CookieJar uses unsafe class caching for dynamically loading cookie jars
+# If 2 rest-client instances are instantiated at the same time, (in threads)
+# non-deterministic behaviour can occur whereby the Hash cookie jar isn't
+# properly loaded and cached.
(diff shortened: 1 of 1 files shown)


bradrobertson:
It looks like it's just Rubocop failures on this test suite (nothing that I actually changed) Is there a standard on what you want done on PRs for this gem? Obviously my changes here aren't the reason this i failing

nickbair-nylas:
@bradrobertson thanks for this!  Don't worry about the RuboCop failures in Travis CI--that's an item that's on our plate to clean up.

nickbair-nylas (approved):
LGTM, thanks!

Jayme-T (approved):

PR #279, opened 2021-03-11
Try parsing all responses as JSON and rescue errors to avoid intermittent parsing errors

ozsivanov:
## Issue
It seems that api.nylas.com fails to include the `application/json` content-type header on some responses. This is causing intermittent parsing errors when assigning values.

## Fixes
Replaced content-type check with logic that always tries parsing a response as JSON, and rescuing if that fails. This should properly parse JSON responses with no content-type header set, and still allow other response types to be passed through un-parsed.

#

### DIFF
--- a/lib/nylas/http_client.rb
+++ b/lib/nylas/http_client.rb
@@ -86,14 +86,12 @@ module Nylas
         timeout: timeout
       )
       rest_client_execute(**request) do |response, _request, result|
-        content_type = nil
-
-        if response.headers && response.headers[:content_type]
-          content_type = response.headers[:content_type].downcase
+        # Try to parse all responses as JSON, since api.nylas.com content-type can be flaky
+        begin
+          response = parse_response(response)
+        rescue Nylas::JsonParseError
         end
-
-        response = parse_response(response) if content_type == "application/json"
-
+        
         handle_failed_response(result: result, response: response)
         response
       end


lib/nylas/http_client.rb
```diff
@@ -86,14 +86,12 @@ def execute(method:, path: nil, headers: {}, query: {}, payload: nil)
...
       )
       rest_client_execute(**request) do |response, _request, result|
-        content_type = nil
-
-        if response.headers && response.headers[:content_type]
-          content_type = response.headers[:content_type].downcase
+        # Try to parse all responses as JSON, since api.nylas.com content-type can be flaky
+        begin
+          response = parse_response(response)
+        rescue Nylas::JsonParseError
```
nickbair-nylas, line 92: @ozsivanov Should we assume that all `Nylas::JsonParseError` are non-JSON responses (files) and therefore pass them on?  I think there would be a case where an exception is raised due to bad JSON.  Is there a way to catch that and not rescue the exception in that case?

nickbair-nylas (changes requested):

lib/nylas/http_client.rb
ozsivanov: I don't think there is a way for us to distinguish between those two cases. I'd like to think that api.nylas.com won't respond with malformed JSON, since we would be currently seeing `Nylas::JsonParseError` getting thrown.

arunagw:
@ozsivanov thanks for the PR. Do you know what all response are missing content/type as not JSON?

mrashed-dev:
Closing as I believe this was addressed in a few PRs like #316

Issue #244, opened 2020-01-07
No specs included for Event model


spang:
Event.rb is entirely untested. This is very sad, and allowed a regression on Event.create() to sneak into the latest release.

arunagw:
Done in #282

PR #282, opened 2021-04-01
Add spec for `Nylas::Event`

arunagw:
This PR adds basic spec for `Nylas::Event` class as these were
missing and also requested in #244

### DIFF
--- a/spec/nylas/event_spec.rb
+++ b/spec/nylas/event_spec.rb
@@ -1,0 +1,126 @@
+# frozen_string_literal: true
+
+describe Nylas::Event do
+  describe ".from_json" do
+    it "Deserializes all the attributes into Ruby objects" do
+      api = instance_double(Nylas::API)
+      data = {
+        id: "event-8766",
+        object: "event",
+        account_id: "acc-1234",
+        busy: true,
+        calendar_id: "cal-0987",
+        description: "an event",
+        message_id: "mess-8766",
+        owner: '"owner" <owner@example.com>',
+        participants: [
+          {
+            comment: "Let me think on it",
+            email: "participant@example.com",
+            name: "Participant",
+            status: "noreply"
+          },
+          {
+            comment: nil,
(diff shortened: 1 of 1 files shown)


spec/nylas/event_spec.rb
```diff
@@ -0,0 +1,152 @@
+# frozen_string_literal: true
+
+describe Nylas::Event do
+  describe ".from_json" do
```
adarsh, line 4: Tiny nit: If it's an instance method, should this be `#from_json`?

spec/nylas/event_spec.rb
```diff
@@ -0,0 +1,152 @@
...
+      api = instance_double(Nylas::API)
+      data = {
+        account_id: "acc-1234",
+        busy: true,
+        calendar_id: "cal-0987"
+      }
+
+      event = described_class.from_json(JSON.dump(data), api: api)
+
+      expect(event.busy?).to eq(true)
```
adarsh, line 82: With RSpec matchers, you should be able to do 

```ruby
expect(event.busy?).to be_true
```
https://relishapp.com/rspec/rspec-expectations/v/2-2/docs/matchers/be-matchers

adarsh (approved):
Looks nice and clean! 👍

spec/nylas/event_spec.rb
arunagw: It's a class method as we are using it like `described_class.from_json`

spec/nylas/event_spec.rb
arunagw: We can even right a better way 

```
expect(event).to be_busy
```

spec/nylas/event_spec.rb
adarsh: Oh whoops, my mistake. There was an instance double so I thought it was an instance variable. Not a great assumption!

PR #283, opened 2021-04-05
Fixes for Ruby 3

grantspeelman:

### DIFF
--- a/gem_config.rb
+++ b/gem_config.rb
@@ -36,6 +36,7 @@ module GemConfig
 
   def self.dev_dependencies
     [["bundler", ">= 1.3.0"],
+     ["mime-types", ">= 3.0.0"],
      ["jeweler", "~> 2.1"],
      ["yard", "~> 0.9.0"],
      ["awesome_print", "~> 1.0"],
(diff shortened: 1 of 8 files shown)


arunagw:
@grantspeelman thanks for the fixes for Ruby 3. We would love to get these changes merged but it seems there are other things failing currently on Ruby 3. Can be seen in #286.

gem_config.rb
```diff
@@ -36,6 +36,7 @@ def self.metadata
 
   def self.dev_dependencies
     [["bundler", ">= 1.3.0"],
+     ["mime-types", ">= 3.0.0"],
```
grantspeelman, line 39: @arunagw made this only a dev dependency minimum as it seems to be only required for ruby 3.
weirdly enough it allows for a later version of jeweler by downgrading a few other gems.

Screenshot_20210407_063934

grantspeelman: maybe worth considering dropping jeweler for the bundler built in gem support

arunagw: I think `jeweler` can be dropped.

grantspeelman: should I drop it in this PR instead of setting minimum mime-types?

arunagw: Yes. I think it's better to drop unused gems.

grantspeelman:
@arunagw good to merge?

arunagw:
Thank you @grantspeelman for the work.

Issue #229, opened 2019-08-01
No way to notify participants on create for events


pk1m:
When creating an event, no way to add query params.
execute takes query params
https://github.com/nylas/nylas-ruby/blob/master/lib/nylas/http_client.rb#L72

but create does not allow it to be passed
https://github.com/nylas/nylas-ruby/blob/master/lib/nylas/model.rb#L43

api should allow us to pass in notify_participants queyr param
https://docs.nylas.com/reference#post-event

rylwin:
@pk1m here's the workaround we're using:

```
        external_event = nylas_client.events.new(params)
        result = external_event.execute(
          method: :post,
          payload: external_event.attributes.serialize,
          path: "#{external_event.resources_path}?notify_participants=true",
        )
```

arunagw:
Closing as #284 is merged in master and this feature will be available in next release. thanks.

PR #284, opened 2021-04-05
Support `notify_participants` on events

arunagw:
We want to support `notify_participants` query param on events
create and update call so when a event created or updated a
notifications can be triggered. This PR enables sending
`notify_participants` in create, update and destroy calls.
This change was also requested in #229

It can be done like:

```
event = nylas.events.new
event.notify_participants = true
event.save
```

Changes:
- Update `Nylas::Model#execute` to support query params
- Update `Nylas::Event` model to support `notify_participants`
  attribute.
- Update `Nylas::Model::Attributes` to handle in case of data is nil.


.rubocop.yml
```diff
@@ -2,7 +2,7 @@ inherit_from: .rubocop_todo.yml
 
 require: rubocop-rspec
 AllCops:
-  TargetRubyVersion: 2.4
+  TargetRubyVersion: 2.5
```
adarsh, line 5: I'm sure you've already considered this but I figured I would ask:

This seems like a minor change but is there any impact on people using this gem with older Ruby versions?

It's only Rubocop also, so I would expect the operational impact to be very minimal but I figured I would ask anyhow, just in case.

lib/nylas/event.rb
```diff
@@ -42,5 +44,15 @@ def rsvp(status, notify_participants:)
...
     end
+
+    private
+
+    def query_params
+      return {} if notify_participants.nil?
+
+      {
+        notify_participants: notify_participants
+      }
```
adarsh, line 55: Unless that explicit `return` is doing something I'm not following, could this also be

```ruby
def query_params
  if notify_participants.nil?
    {}
  else
    { notify_participants: notify_participants }
  end
end
```
?

lib/nylas/model.rb
```diff
@@ -99,6 +109,10 @@ def save_call
       )
     end
 
+    def query_params
+      {}
```
adarsh, line 113: If you're initializing `query_params` as `{}` by default in line 42, do you still need this method?

lib/nylas/model/attributes.rb
```diff
@@ -23,6 +23,7 @@ def []=(key, value)
 
       # Merges data into the registry while casting input types correctly
       def merge(new_data)
+        new_data ||= {}
```
adarsh, line 26: Is it possible to create this default assignment in the constructor? 

```ruby
def merge(new_data = {})
  # ...
end

adarsh (approved):
Minor syntax questions from me, looks great!

lib/nylas/model/attributes.rb
arunagw: Removed here 36d840047b13ce05090484b4d7936e1eb1edca8f as it's not needed. Thanks

lib/nylas/event.rb
arunagw: Fixed here 36d840047b13ce05090484b4d7936e1eb1edca8f

lib/nylas/model.rb
arunagw: Yes we need to as this `Model` class is used in other places like events and message where we can override this.

Also `query_params` is used https://github.com/nylas/nylas-ruby/blob/support-notify-participants-on-event/lib/nylas/model.rb#L53

.rubocop.yml
arunagw: I think it's fine to change this as it's more of our internal. Also we have dropped support of 2.4 so it's fine to use 2.5 here.

Issue #258, opened 2020-10-12
Cannot download a file that was fetched from a message association


bmulholland:
Calling download on a file that was fetched from a message fails. This is common if you're looping over a set of messages and then downloading the files from there.

Workaround is to do a fresh file fetch.

This code fails:
```
2.7.2 :021 > nylas.messages.find('ID_FOR_MESSAGE_WITH_ATTACHMENT').files.first.download
Traceback (most recent call last):
        2: from (irb):20
        1: from (irb):21:in `rescue in irb_binding'
NoMethodError (undefined method `get' for nil:NilClass)
Did you mean?  gem
```

This works, but is more complex and less efficient:
```
2.7.2 :023 > nylas.files.find(message.files.first.id).download
 => #<Tempfile:/var/folders/vz/wfs0t4g55xn0sjpnmf6d9jfc0000gn/T/Services Contract - Hellsgate - Reply With Redlines Accepted.docx20201012-70192-ypr2bz>
```

arunagw:
I can confirm this bug.

arunagw:
@bmulholland this has been fixed in main branch and will be available in next release. Thanks for reporting this.

PR #287, opened 2021-04-14
Fix issue where download a file fetch via `find`

arunagw:
This PR fixes issue where downloading a file for the
message which is fetched via `find_model` method.
This PR updates `find_model` to fetch from API and then use
`from_hash` to initialize object instead of just using `reload`
which was internally using `assign` method to assign values.
By doing this way `file` object properly gets `api` attribute
assigned and `download` method works.
This issue was reported in #258

Changes:
- Update `Nylas::Collection` to update `find_model` method
  to fetch again fron API and use `from_hash` method.

### DIFF
--- a/lib/nylas/collection.rb
+++ b/lib/nylas/collection.rb
@@ -127,8 +127,13 @@ module Nylas
     end
 
     def find_model(id)
-      instance = model.from_hash({ id: id }, api: api)
-      instance.reload
+      response = api.execute(
+        **to_be_executed.merge(
+          path: "#{resources_path}/#{id}",
+          query: {}
+        )
+      )
+      instance = model.from_hash(response, api: api)
       instance
     end
 
--- a/lib/nylas/collection.rb
+++ b/lib/nylas/collection.rb
@@ -133,8 +133,7 @@ module Nylas
           query: {}
         )
       )
-      instance = model.from_hash(response, api: api)
-      instance
+      model.from_hash(response, api: api)
     end
 
     # @return [Hash] Specification for request to be passed to {API#execute}
--- a/lib/nylas/collection.rb
+++ b/lib/nylas/collection.rb
@@ -128,7 +128,7 @@ module Nylas
 
     def find_model(id)
       response = api.execute(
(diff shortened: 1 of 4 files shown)


lib/nylas/collection.rb
```diff
@@ -127,9 +127,13 @@ def resources_path
     end
 
     def find_model(id)
-      instance = model.from_hash({ id: id }, api: api)
-      instance.reload
-      instance
+      response = api.execute(
+        **to_be_executed.merge(
```
adarsh, line 131: I believe the double splat and `.merge` might be redundant here, but I could be wrong. Does it work without the splats?

Reference: https://jetrockets.pro/blog/ruby-double-splat-operator-cheatsheet

spec/spec_helper.rb
```diff
@@ -45,6 +45,11 @@ class FullModel
   attribute :web_page, :web_page
 
   has_n_of_attribute :web_pages, :web_page
+  has_n_of_attribute :files, :file
+
+  attr_accessor :api
```
adarsh, line 50: Does this make `:api` accessible to all `FullModel` classes for the tests? Is there a way we can limit it's scope to where we need it for the tests only? 

This is a very minor note, focused on minimizing public surface area, feel free to ignore.

adarsh (commented):
👍 Nice solution. This will be helpful once fixed!

philrenaud (approved):

spec/spec_helper.rb
arunagw: I think we need to because `FullModel` is used everywhere in test and we needed to recreate the scenario like we are doing in real models.

lib/nylas/collection.rb
arunagw: We need to use it here to support with ruby 3.0 https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

adarsh: Oh nice!

PR #288, opened 2021-04-16
Add support for `/free-busy` endpoint

arunagw:
This PR adds support for `/calendars/free-busy` endpoint to get
information about free/busy inforamtion for a calendar during a
certain time period.

This PR adds `FreeBusyCollection` to handle making post call to
server and to create object of `FreeBusy` and `TimeSlot` from the
response. This PR also adds `FreeBusy` and `TimeSlot` class to handle
objects representation respectivily.

Example:
```
api = Nylas::API.new(
    app_id: ENV['NYLAS_APP_ID'],
    app_secret: ENV['NYLAS_APP_SECRET'],
    access_token: ENV['NYLAS_ACCESS_TOKEN']
    )

result = api.free_busy(emails: emails, start_time: start_time, end_time: end_time)

free_busy = result.first
free_busy.object #=> 'free_busy'
free_busy.email #=> 'given email address'
free_busy.time_slots #=> Object of `TimeSlot` with information
```

Documentaion:
https://docs.nylas.com/reference#calendars-free-busy


lib/nylas/api.rb
```diff
@@ -150,6 +150,15 @@ def webhooks
       @webhooks ||= Collection.new(model: Webhook, api: as(client.app_secret))
     end
 
+    def free_busy(emails:, start_time:, end_time:)
+      FreeBusyCollection.new(
+        api: self,
+        emails: emails,
+        start_time: start_time.to_i,
```
adarsh, line 157: You probably already know but there aren't any timezone issues here, are there? What format does `start_time` come in (e.g. `DateTime`)?

lib/nylas/free_busy_collection.rb
```diff
@@ -0,0 +1,47 @@
...
+      return enum_for(:each) unless block_given?
+
+      execute.each do |result|
+        yield(FreeBusy.new(**result))
+      end
+    end
+
+    private
+
+    attr_reader :api, :emails, :start_time, :end_time
```
adarsh, line 27: You know I love the private `attr_reader` 💕

lib/nylas/free_busy_collection.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    def execute
+      api.execute(
+        method: :post,
+        path: "/calendars/free-busy",
```
adarsh, line 32: Should this be a constant?

lib/nylas/time_slot.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module Nylas
+  # Query free/busy information for a calendar during a certain time period
+  # @see https://docs.nylas.com/reference#calendars-free-busy
+  class TimeSlot
+    include Model::Attributable
+
+    attribute :object, :string
+    attribute :status, :string
+    attribute :start_time, :unix_timestamp
```
adarsh, line 11: Oh these are Unix timestamps so TZ is not an issue. 👍

spec/nylas/api_spec.rb
```diff
@@ -56,6 +56,59 @@
...
 
+  describe "#free_busy" do
+    it "returns `Nylas::FreeBusyCollection` for free busy details" do
+      emails = ["test@example.com", "anothertest@example.com"]
+      start_time = 1_609_439_400
+      end_time = 1_640_975_400
+      client = Nylas::HttpClient.new(
+        app_id: "not-real",
+        app_secret: "also-not-real",
+        access_token: "seriously-unreal"
```
adarsh, line 67: 🤣

spec/nylas/api_spec.rb
```diff
@@ -56,6 +56,59 @@
...
+              status: "busy",
+              start_time: 1_609_439_400,
+              end_time: 1_640_975_400
+            }
+          ]
+        }
+      ]
+      allow(client).to receive(:execute).with(
+        method: :post,
+        path: "/calendars/free-busy",
```
adarsh, line 86: If you ended up making the path a constant, you could refer to the constant directly here, maybe the test is slightly less brittle then?

spec/nylas/free_busy_collection_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      allow(api).to receive(:execute).and_return(free_busy_response)
+      result = described_class.new(
+        api: api,
+        emails: emails,
+        start_time: start_time.to_i,
+        end_time: end_time.to_i
+      ).to_a
+
+      expect(result.size).to eq(2)
+      first_result = result.first
```
adarsh, line 40: Is there a reason not to iterate over all of the objects in this array and test them? I'm assuming performance?

spec/nylas/api_spec.rb
```diff
@@ -56,6 +56,59 @@
...
+      ).and_return(free_busy_response)
+
+      result = api.free_busy(
+        emails: emails,
+        start_time: Time.at(start_time),
+        end_time: Time.at(end_time)
+      )
+
+      expect(result).to be_a(Nylas::FreeBusyCollection)
+      free_busy = result.last
```
adarsh, line 101: Is there a reason not to iterate over all of the objects in this array and test them? I'm assuming performance?

spec/nylas/free_busy_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+          status: "busy",
+          start_time: 1_609_439_400,
+          end_time: 1_640_975_400
+        }
+      ]
+
+      result = described_class.new(time_slots: time_slots)
+
+      expect(result.time_slots.to_a.size).to eq(1)
+      time_slot = result.time_slots.last
```
adarsh, line 40: Same about iterating over all the objects. Alternative: test that there is only one `time_slot` object in the collection then leave this line. When you have a `.last` and only test that one object, it makes me wonder about the others in the collection.

Not a huge deal at all, just wondering.

adarsh (approved):
This looks excellent - it lines up nicely with the API documentation.

I have some minor questions, otherwise LGTM.

lib/nylas/api.rb
arunagw: No issues as it's unix timestamp

spec/nylas/api_spec.rb
arunagw: I would keep the real string here in the test but using constant in code d7a0a721b0a6afe0b62a111eaf0692d14d732439

spec/nylas/free_busy_collection_spec.rb
arunagw: We just have one object so we don't really need to loop over things here.

spec/nylas/api_spec.rb
arunagw: First and last are different as one have time_slots and other do not have.

spec/nylas/free_busy_spec.rb
arunagw: The array contains only one so `.last` or `.first` is fine I think.

Issue #120, opened 2017-11-09
Make certain parameters "read only" and make them raise an error if they are changed


zspencer:
Certain attributes, such as `name` on folders and many of the message attributes are read-only, however they can be mutated by an SDK user.

Let's provide them with some user-friendly feedback when attempting to set those parameters (or perhaps make their setters private or protected?)

arunagw:
Done in #298

PR #291, opened 2021-04-22
Exclude `id`, `object` and `account_id` when updating an event

arunagw:
We have identified that few of the event update calls to the API
sends `object` and `account_id` attributes and it should not be
allowed to send.
This PR stops sending `id`, `object` and `account_id` attribute
to the API when we call `.save` on an event object. These are
readonly object on server and we should avoid sending it.

### DIFF
--- a/lib/nylas/event.rb
+++ b/lib/nylas/event.rb
@@ -9,9 +9,9 @@ module Nylas
     allows_operations(creatable: true, listable: true, filterable: true, showable: true, updatable: true,
                       destroyable: true)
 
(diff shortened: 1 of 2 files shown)


arunagw:
This will prevent errors on API side for now but in future I think we should add something suggested in #120

adarsh (approved):
Very straightforward. 👍🏾 

Consider linking to the section in the API docs in the merge commit in case anyone wants to reference it later - not a huge thing though.

PR #292, opened 2021-04-22
Add support for authentication_url for hosted authentication

arunagw:
This PR adds `authentication_url` method to generate a URL to redirect
to login through Nylas hosted authentication flow.

Mode information can be found here about how authentication flow works
https://docs.nylas.com/reference#oauth

These changes also requested in #223 and also worked on a bit in #227

Changes:
- Update `Nylas::API` to add `authentication_url` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-## 4.6.7 / 2021-04-22
+### Unreleased
+
+* Add support for hosted authentication #292
+
+### 4.6.7 / 2021-04-22
 
 * Support for Ruby 3.
 * Add support for `/free-busy` endpoint #288
--- a/lib/nylas/api.rb
+++ b/lib/nylas/api.rb
@@ -38,16 +38,14 @@ module Nylas
     end
 
     def authentication_url(redirect_uri:, scopes:, response_type: "code", login_hint: nil, state: nil)
-      scopes = scopes.join(",") if scopes
-
       params = {
         client_id: app_id,
         redirect_uri: redirect_uri,
         response_type: response_type,
-        scopes: scopes,
         login_hint: login_hint
       }
       params[:state] = state if state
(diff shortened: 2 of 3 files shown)


arunagw:
Should we call it `url_for_authentication` instead of `authentication_url` as it used to be called `url_for_authentication` in older version of gem based on this commit https://github.com/nylas/nylas-ruby/commit/c6617d42780a1deb60f240c7495598b1406edae0

CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
-## 4.6.7 / 2021-04-22
+### Unreleased
```
adarsh, line 1: 👍🏾 This is a great idea

lib/nylas/api.rb
```diff
@@ -37,6 +37,21 @@ def authenticate(name:, email_address:, provider:, settings:, reauth_account_id:
       )
     end
 
+    def authentication_url(redirect_uri:, scopes:, response_type: "code", login_hint: nil, state: nil)
+      scopes = scopes.join(",") if scopes
```
adarsh, line 41: Does this fail if it's `nil` and you don't add the check? Just curious.

lib/nylas/api.rb
```diff
@@ -37,6 +37,21 @@ def authenticate(name:, email_address:, provider:, settings:, reauth_account_id:
...
+
+      params = {
+        client_id: app_id,
+        redirect_uri: redirect_uri,
+        response_type: response_type,
+        scopes: scopes,
+        login_hint: login_hint
+      }
+      params[:state] = state if state
```
adarsh, line 50: Do we not want to include this key if it's nil?

Is this simpler if we just include the key `state: nil` and handle that downstream?

adarsh (approved):
Minor questions, LGTM ✨

PR #293, opened 2021-04-29
Support for `messages.expanded.find(id)` to get expanded message

### LINKED ISSUE #189: Support expanded view when using find on messages

arunagw:
Currently sending `view` in the query param is not supported when
fetching a single record and we want to support it as it's already
supported in API. This PR allows to chain `view` related methods before
`find` so that `view` is send to API.

This PR fixes #189 where it has been requested that expanded message
should be fetched in single call by sending `view=expanded` in query param.

Changes:
- Update `Nylas::Collection` to send `view` in query params when a view
related method is chained.

### DIFF
--- a/spec/nylas/collection_spec.rb
+++ b/spec/nylas/collection_spec.rb
@@ -74,7 +74,7 @@ describe Nylas::Collection do
       expect(instance.api).to eq(api)
     end
 
-    it "retrieves a single object with `view` argument in query if clauses earlier in the chain" do
+    it "retrieves with `view` argument in query if clauses earlier in the chain" do
       collection = described_class.new(model: FullModel, api: api)
       allow(api).to receive(:execute).and_return(example_instance_hash)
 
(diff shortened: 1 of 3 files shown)


spec/nylas/collection_spec.rb
```diff
@@ -74,6 +74,22 @@ def example_instance_hash
       expect(instance.api).to eq(api)
     end
 
+    it "retrieves a single object with `view` argument in query if clauses earlier in the chain" do
```
adarsh, line 77: Since we have an `if` statement, is it worth testing the other branch of the conditional?

adarsh (approved):
Minor question, short and sweet 🍬

spec/nylas/collection_spec.rb
arunagw: Added here https://github.com/nylas/nylas-ruby/pull/293/commits/9b6069fdadfb539130683f7d4bf0a637d0e1ae2c

PR #296, opened 2021-05-06
Add `Nylas-API-Version` header to be sent to API

arunagw:
Based on versioning docs https://developer.nylas.com/docs/developer-tools/api/versioning/
we should be sending `Nylas-API-Version` header with supprted version
number to the API. This PR impements it and also set it to 2.2 which is
latest version of API

Changes:
- Update `Nylas::HttpClient` to send `Nylas-API-Version` header.

### DIFF
--- a/spec/nylas/http_client_spec.rb
+++ b/spec/nylas/http_client_spec.rb
@@ -45,13 +45,14 @@ describe Nylas::HttpClient do
 
   describe "#execute" do
     it "includes Nylas API Version in headers" do
+      supported_api_version = "2.2"
       nylas = described_class.new(app_id: "id", app_secret: "secret", access_token: "token")
       allow(RestClient::Request).to receive(:execute)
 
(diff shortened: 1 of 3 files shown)


spec/nylas/http_client_spec.rb
```diff
@@ -43,6 +43,23 @@
...
 
+  describe "#execute" do
+    it "includes Nylas API Version in headers" do
+      nylas = described_class.new(app_id: "id", app_secret: "secret", access_token: "token")
+      allow(RestClient::Request).to receive(:execute)
+
+      nylas.execute(method: :get, path: "/contacts/1234/picture")
+
+      expect(RestClient::Request).to have_received(:execute).with(
+        headers: hash_including("Nylas-API-Version" => "2.2"),
```
adarsh, line 54: Consider replacing "2.2" with a reference to the constant in the class under test - probably will be less brittle.

adarsh (approved):
👍🏾 one minor suggestion

Issue #266, opened 2020-12-22
account_id nil on delta converted to message model


sudall:
Hello,
Looks like there is a bug here when converting a delta to a message model:

https://github.com/nylas/nylas-ruby/blob/7babaebf3866d256f190b2e0d45858d12483b379/lib/nylas/delta.rb#L28

There is no `account_id` property on the create message delta json as you can see in the examples in the docs: https://docs.nylas.com/docs/deltas-overview#deltas-response-examples. So, when merging account_id onto object_attributes, account_id gets set to nil on the message model. account_id is not nil on object_attributes, however; so it seems that attribute should be used.

arunagw:
Hey @sudall thanks for reporting it. I was looking at the docs and found that response do have an `account_id` under `deltas`. Can you please check and see if this still an issue for you? Also if it's still an issue please provide a sample code or snippet to reproduce this. thanks.

arunagw:
Closing this for now. Please reopen this if still an issue. Thanks.

sudall:
Apologies if this was unclear, but I believe this is still an issue. There is no `account_id` property for deltas one level deep into the object (only `cursor`, `event`, `id`, and `object`). There is an `account_id` on the `attributes` property (which appears to be represented by `object_attributes` in the Delta model).

arunagw:
OK I can see this now. I am going to reopen this and will try to replicate the problem and will revert back. thanks

arunagw:
We have merged #297 to fix problems with delta. Can you please check with latest master and see if things are working ok? thanks

arunagw:
This has been taken care in #297

PR #297, opened 2021-05-11
Fix attributes assignment for `Delta`

arunagw:
Background:
Currently `Delta` object for changes like `Message` and `Event`
having `account_id` as nil and it was also reported in #266.
This PR fixes the issue where the delta object properly assigning
`account_id` and other attributes to the attached model object
for messages and others.

Changes:
- Update `Delta` class to handle case where changes are inside
  `attributes` key. Also marked `object_attributes_with_ids` as
  private.
- Update `Deltas` class to have `.last` method which useful when
  having multiple objects.

### DIFF
--- a/lib/nylas/delta.rb
+++ b/lib/nylas/delta.rb
@@ -6,6 +6,7 @@ module Nylas
   # @see https://docs.nylas.com/reference#deltas
   class Delta
     include Model::Attributable
+
     attribute :id, :string
     attribute :type, :string
     attribute :object, :string
@@ -42,8 +43,10 @@ module Nylas
              else
                data
              end
+
+      data = data.merge(data[:attributes]) if data[:attributes]
       data[:object_attributes] = data.delete(:attributes)
-      super(data)
+      super(**data)
     end
   end
 end
--- a/lib/nylas/delta.rb
+++ b/lib/nylas/delta.rb
@@ -25,8 +25,6 @@ module Nylas
       @model ||= Types.registry[object.to_sym].cast(object_attributes_with_ids)
     end
 
-    private
-
     def object_attributes_with_ids
       (object_attributes || {}).merge(id: id, account_id: account_id)
     end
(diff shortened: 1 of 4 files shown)


lib/nylas/delta.rb
```diff
@@ -42,8 +45,10 @@ def cast(data)
              else
                data
              end
+
+      data = data.merge(data[:attributes]) if data[:attributes]
```
adarsh, line 49: Random question: What would happen if you merged a nil or missing hash?

spec/nylas/delta_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    expect(delta.object).to eq("message")
+    expect(delta.account_id).to eq("acc-id")
+    expect(delta.id).to eq("message-id")
+    expect(delta.type).to eq("message")
+    expect(delta.event).to eq("created")
+    expect(delta.cursor).to eq("999")
+    expect(delta.namespace_id).to eq("namespace-id")
+    expect(delta.date).to eq(Time.at(1_609_439_400))
+    expect(delta.metadata).to eq(key: :value)
+    expect(delta.object_attributes).to eq(object: :attributes)
```
adarsh, line 29: If you have this more comprehensive test, do you need the other ones also?

spec/nylas/deltas_spec.rb
```diff
@@ -44,4 +82,59 @@
     expect(delta.model.thread_id).to eql "thread-098"
     expect(delta.model.received_date).to eql Time.at(1_514_339_665)
   end
+
+  it "parse stream data from multiple changes" do
```
adarsh, line 86: Typos? "parses"?

adarsh (approved):
👍🏽

spec/nylas/delta_spec.rb
arunagw: I thought of covering every attribute we have defined in the class in one test. But May be we can split this into multiples so it reads better.

lib/nylas/delta.rb
arunagw: If it's `nil` it will raise an error and that's why we have `if` condition. But if it's empty it will do nothing. I will try to add test for these condition as well so it will not break.

PR #298, opened 2021-05-14
Support for read only attributes

arunagw:
Background:
Currently in API there are few attributes which should be treated as
read only and we should not send them to API when creating and updating.
We did a solution in #291 but it's not working for the cases where
attributes are nested. This change also suggested in #120.

This PR allows us setting up read only attibute on attribute level
and they will not be sent to the API when creating and updating.

Changes:
- Removed usage of `exclude_when` and using `read_only` to determine
  if attribute is read_only or not.
- Added `serialize_for_api` method in `Nylas::Model::Attributes` to
  be used for create and update calls.

### DIFF
--- a/lib/nylas/contact.rb
+++ b/lib/nylas/contact.rb
@@ -13,9 +13,9 @@ module Nylas
     self.updatable = true
     self.destroyable = true
 
-    attribute :id, :string, exclude_when: %i[creating updating]
+    attribute :id, :string, read_only: true
     attribute :object, :string, default: "contact"
-    attribute :account_id, :string, exclude_when: %i[creating updating]
+    attribute :account_id, :string, read_only: true
 
     attribute :given_name, :string
     attribute :middle_name, :string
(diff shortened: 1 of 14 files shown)


lib/nylas/contact.rb
```diff
@@ -13,9 +13,9 @@ class Contact
     self.updatable = true
     self.destroyable = true
 
-    attribute :id, :string, exclude_when: %i[creating updating]
+    attribute :id, :string, read_only: true
```
adarsh, line 16: This is much clearer than the excludes 👍🏽

lib/nylas/model/attributes.rb
```diff
@@ -39,6 +39,12 @@ def serialize(keys: attribute_definitions.keys)
         JSON.dump(to_h(keys: keys))
       end
 
+      def serialize_for_api(keys: attribute_definitions.keys)
+        api_keys = keys.delete_if { |key| attribute_definitions[key].read_only == true }
```
adarsh, line 43: Very nice.

adarsh (approved):
Elegant!

PR #300, opened 2021-05-22
[58503] Ruby SDK support for metadata for Events

mrashed-dev:
# Description
The Nylas Calendar API has a new beta feature of adding a new [metadata field to calendar events](https://www.nylas.com/blog/event-metadata). This PR makes the feature available via the Nylas Ruby SDK. 

# Usage
To create a new event with event metadata, you can create a `Hash` object with a mapping of key-value pairs for the metadata:
```
api.events.create(
  title: "A fun event!",
  ...
  metadata: {
    event_type: "gathering"
  }
).to_h
```

To query events based on metadata, you can filter on three different parameters:
- `metadata_key` (string or array) to filter based on the keys within the metadata object
- `metadata_value` (string or array) to filter based on the value within the metadata object
- `metadata_pair` (hash) to filter based on the key-value pairs within the metadata object

These filters can be passed in as parameters, like so:
- `events = calendar.events.where(metadata_key: "event_type")`
- `events = calendar.events.where(metadata_value: ["gathering", "meeting"])`
- `events = calendar.events.where(metadata_pair: {event_type: "gathering"})`

# License

### DIFF
--- a/lib/nylas/event.rb
+++ b/lib/nylas/event.rb
@@ -27,6 +27,7 @@ module Nylas
     attribute :status, :string
     attribute :title, :string
     attribute :when, :when
+    attribute :metadata, :hash
     attribute :original_start_time, :unix_timestamp
 
     attr_accessor :notify_participants
--- a/lib/nylas/http_client.rb
+++ b/lib/nylas/http_client.rb
@@ -104,8 +104,14 @@ module Nylas
                         also_log: { result: true, values: %i[method url path headers query payload] }
 
     def build_request(method:, path: nil, headers: {}, query: {}, payload: nil, timeout: nil)
-      headers[:params] = query
       url ||= url_for_path(path)
+
+      # Add parameters using the uri module as it allows us to make custom rules
+      uri = URI.parse(url)
+      params = URI.decode_www_form(uri.query || '') + query.to_a
+      uri.query = URI.encode_www_form(params)
+      url = uri.to_s
+
       resulting_headers = default_headers.merge(headers)
       { method: method, url: url, payload: payload, headers: resulting_headers, timeout: timeout }
     end
--- a/lib/nylas/http_client.rb
+++ b/lib/nylas/http_client.rb
@@ -184,7 +184,6 @@ module Nylas
       raise exception.new(response[:type], response[:message], response.fetch(:server_error, nil))
     end
 
-    # Add parameters using the uri module as it allows us to handle special cases
     def add_query_params_to_url(url, query)
       unless query.empty?
         uri = URI.parse(url)
@@ -197,15 +196,15 @@ module Nylas
       url
     end
 
-    # Process any URL parameter that requires a special case
     def custom_params(query)
+      # Convert hash to "<key>:<value>" form for metadata_pair query
       if query.key?(:metadata_pair)
-        pairs = []
-        query[:metadata_pair].each do |key, value|
-          pairs.append("#{key}:#{value}")
+        pairs = query[:metadata_pair].map do |key, value|
+          "#{key}:#{value}"
         end
         query[:metadata_pair] = pairs
       end
+
       query
     end
   end
--- a/lib/nylas/http_client.rb
+++ b/lib/nylas/http_client.rb
@@ -105,16 +105,7 @@ module Nylas
 
     def build_request(method:, path: nil, headers: {}, query: {}, payload: nil, timeout: nil)
       url ||= url_for_path(path)
-
-      # Add parameters using the uri module as it allows us to make custom rules
-      unless query.empty?
-        uri = URI.parse(url)
-        query = custom_params(query)
-        params = URI.decode_www_form(uri.query || "") + query.to_a
-        uri.query = URI.encode_www_form(params)
-        url = uri.to_s
-      end
-
+      url = add_query_params_to_url(url, query)
       resulting_headers = default_headers.merge(headers)
       { method: method, url: url, payload: payload, headers: resulting_headers, timeout: timeout }
     end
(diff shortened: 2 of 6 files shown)


lib/nylas/http_client.rb
```diff
@@ -104,8 +104,17 @@ def execute(method:, path: nil, headers: {}, query: {}, payload: nil)
                         also_log: { result: true, values: %i[method url path headers query payload] }
 
     def build_request(method:, path: nil, headers: {}, query: {}, payload: nil, timeout: nil)
-      headers[:params] = query
       url ||= url_for_path(path)
+
+      # Add parameters using the uri module as it allows us to make custom rules
+      unless query.empty?
```
arunagw, line 110: We need to add some tests for this to make sure we are not making anything wrong here.

lib/nylas/http_client.rb
```diff
@@ -104,8 +104,17 @@ def execute(method:, path: nil, headers: {}, query: {}, payload: nil)
                         also_log: { result: true, values: %i[method url path headers query payload] }
 
     def build_request(method:, path: nil, headers: {}, query: {}, payload: nil, timeout: nil)
-      headers[:params] = query
       url ||= url_for_path(path)
+
+      # Add parameters using the uri module as it allows us to make custom rules
```
arunagw, line 109: We need to add some tests for this to make sure we are not making anything wrong here.

lib/nylas/http_client.rb
```diff
@@ -104,8 +104,17 @@ def execute(method:, path: nil, headers: {}, query: {}, payload: nil)
...
+      unless query.empty?
+        uri = URI.parse(url)
```
arunagw, line 111: We need to add some tests for this to make sure we are not making anything wrong here.

arunagw (changes requested):

lib/nylas/event.rb
```diff
@@ -27,6 +27,7 @@ class Event
     attribute :status, :string
     attribute :title, :string
     attribute :when, :when
+    attribute :metadata, :hash
```
arunagw, line 30: Should we make a new type called `Metadata` or it can contain anything `key`, `value` pairs?

arunagw:
Hey @mrashed-dev thanks for the PR. Great work.

lib/nylas/http_client.rb
mrashed-dev: Good call out! Added test cases for building different possible query params!

lib/nylas/event.rb
mrashed-dev: `Metadata` is just a bunch of string key-value pairs. It can be anything the user wants it to be, it's abstract in a sense. I think we should leave it as is instead of creating a type for it.

lib/nylas/http_client.rb
mrashed-dev: Good idea! It's implemented now!

spec/nylas/http_client_spec.rb
```diff
@@ -90,4 +90,45 @@
...
+      nylas = described_class.new(app_id: "id", app_secret: "secret", access_token: "token")
+      request = nylas.build_request(method: :get, path: path, query: {})
+      expect(CGI.unescape(request[:url])).to eql(url + path)
+    end
+
+    it "one query parameter" do
+      nylas = described_class.new(app_id: "id", app_secret: "secret", access_token: "token")
+      request = nylas.build_request(method: :get, path: path, query: { param: "value" })
+      expected_params = "?param=value"
+      expect(CGI.unescape(request[:url])).to eql(url + path + expected_params)
```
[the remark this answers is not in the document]
mrashed-dev: We can do that for all cases but the last case, the `:` symbol gets urlencoded. We can approach this in 3 ways:

- Directly write `%3A` for that one case to compare the strings without the use of `CGI.unescape` (e.g. `?metadata_pair=key1%3Avalue1&metadata_pair=key2%3Avalue2&metadata_pair=key3%3Avalue3`
- Remove CGI.unescape for the first 4 cases and keep it for the one that needs it
- Keep it as is for consistency

Which one do you feel strongest about?

lib/nylas/http_client.rb
```diff
@@ -183,5 +192,17 @@ def handle_anticipated_failure_mode(http_code:, response:)
...
+    def custom_params(query)
+      if query.key?(:metadata_pair)
+        pairs = []
+        query[:metadata_pair].each do |key, value|
+          pairs.append("#{key}:#{value}")
+        end
+        query[:metadata_pair] = pairs
+      end
+      query
+    end
```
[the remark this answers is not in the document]
mrashed-dev: Thanks! Fixed.

spec/nylas/http_client_spec.rb
```diff
@@ -90,4 +90,45 @@
...
   end
+
+  describe "building URL with query params" do
+    url = "https://token:@api.nylas.com"
+    path = "/contacts/1234/picture"
+
+    it "no query parameters" do
+      nylas = described_class.new(app_id: "id", app_secret: "secret", access_token: "token")
+      request = nylas.build_request(method: :get, path: path, query: {})
+      expect(CGI.unescape(request[:url])).to eql(url + path)
```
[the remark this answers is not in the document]
mrashed-dev: Interesting, thanks for providing me the link. I think I updated it to the correct way, can you just double check I didn't miss anything?

spec/nylas/http_client_spec.rb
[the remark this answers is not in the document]
arunagw: Let's keep it as it is.

arunagw (approved):
Looks good! Let's squash and merge. 🚀

PR #302, opened 2021-05-26
Fix issue where files couldn't be attached to drafts

mrashed-dev:
# Description
When attaching a file, the user would pass in the `File` object after a successful upload, however without extracting the `file_ids` the draft would be created without attachments

# License

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ Gemfile.lock
 gemfiles/*.lock
 InstalledFiles
 .idea/
+*.iml
 _yardoc
 coverage
 doc/
--- a/lib/nylas/draft.rb
+++ b/lib/nylas/draft.rb
@@ -40,8 +40,24 @@ module Nylas
 
     def create
       unless files.nil? || files.empty?
-        f = files.map(&:id)
-        self.file_ids = f
+        extract_file_ids
+      end
+
+      super
+    end
+
+    def save_call
+      unless files.nil? || files.empty?
+        extract_file_ids
+      end
+
+      super
+    end
+
+    def update(**data)
+      unless files.nil? || files.empty?
+        extract_file_ids
+        data[:file_ids] = self.file_ids;
       end
 
       super
@@ -64,6 +80,13 @@ module Nylas
 
     def destroy
       execute(method: :delete, path: resource_path, payload: attributes.serialize_for_api(keys: [:version]))
+    end
+
+    private
+
+    def extract_file_ids
+      f = files.map(&:id)
+      self.file_ids = f
     end
   end
 end
--- a/lib/nylas/draft.rb
+++ b/lib/nylas/draft.rb
@@ -38,15 +38,7 @@ module Nylas
 
     transfer :api, to: %i[events files folder labels]
 
-    def create
-      extract_file_ids unless files.nil? || files.empty?
-      super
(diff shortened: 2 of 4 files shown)


lib/nylas/draft.rb
```diff
@@ -30,13 +30,23 @@ class Draft
...
     has_n_of_attribute :files, :file
+    has_n_of_attribute :file_ids, :string
     attribute :folder, :folder
     has_n_of_attribute :labels, :label
 
     attribute :tracking, :message_tracking
 
     transfer :api, to: %i[events files folder labels]
 
+    def create
```
arunagw, line 41: Do we also need to support this for update call?

I am also thinking of making `files` attribute as read_only so it won't get send to server. I think currently setting `files` with array of file object will send `files` key to server and API won't accept it and it might return error in future.

mrashed-dev: Good call on both. 
- `files` is now read-only
- Added the extracting `file_id` to both save and update functions

lib/nylas/draft.rb
```diff
@@ -29,14 +29,34 @@ class Draft
...
+      extract_file_ids unless files.nil? || files.empty?
+      super
+    end
+
+    def save_call
```
arunagw, line 46: This can be private I think

lib/nylas/draft.rb
```diff
@@ -29,14 +29,34 @@ class Draft
     attribute :unread, :boolean
 
     has_n_of_attribute :events, :event
-    has_n_of_attribute :files, :file
+    has_n_of_attribute :files, :file, read_only: true
+    has_n_of_attribute :file_ids, :string
```
arunagw, line 33: We also need to add few tests for this new custom code.

lib/nylas/draft.rb
```diff
@@ -29,14 +29,34 @@ class Draft
...
+      super
+    end
+
+    def save_call
+      extract_file_ids unless files.nil? || files.empty?
+      super
+    end
+
+    def update(**data)
+      unless files.nil? || files.empty?
```
arunagw, line 52: I think this condition can also be moved into a private method. As it's used in `create` and `update`

lib/nylas/draft.rb
mrashed-dev: Yup, tested and confirmed. Change is made.

lib/nylas/draft.rb
mrashed-dev: Good spot, moved into the `extract_file_ids` method.

lib/nylas/draft.rb
mrashed-dev: Added a new test to check for this and the functionality of extracting `file_ids` from `files`. Let me know if it's good enough!

.gitignore
```diff
@@ -9,6 +9,7 @@ Gemfile.lock
 gemfiles/*.lock
 InstalledFiles
 .idea/
+*.iml
```
adarsh, line 12: Any idea what this is?

adarsh (approved):
LGTM!

.gitignore
mrashed-dev: This is for blacklisting the Intellij IDEA IDE cache file

Issue #274, opened 2021-02-05
Methods that should only return ``Collection`` type return ``RestClient::Response`` instead of raising exceptions on 502


moigagoo:
Methods that we expect to return only [collections](https://github.com/nylas/nylas-ruby/blob/main/lib/nylas/collection.rb), e.g. api.threads.search(params).execute, also return ``RestClient::Response`` in case of error.

It would be more convenient for us to get an exception in this case because the way it's done now, we can't reliably process the response and have to expect ``RestClient::Response`` just in case.

arunagw:
@moigagoo I would say it should return errors with `Nylas` namespace instead of `RestClient::Response` as suggested in #272 

What do you think?

moigagoo:
@arunagw if by returning errors you mean raising exceptions, then yes, this is exactly what I'm asking for. The exception should belong to the Nylas namespace, of course.

arunagw:
@moigagoo can you please give us a code snippet or example code to try this out? I can see that error raised in https://github.com/nylas/nylas-ruby/blob/main/lib/nylas/http_client.rb#L177-L183 for the known errors.

arunagw:
This has been fixed by #303. Please give it a try if you can. 

thanks

Issue #272, opened 2021-01-15
Request timeout errors should be wrapped with Nylas namespace


nickbair-nylas:
When a request times out, the error that is raised is not part of the Nylas namespace.  This makes rescuing the error difficult for users of the gem.

`Timed out reading data from server (RestClient::Exceptions::ReadTimeout)`

bmulholland:
This is a nice quality-of-life fix that makes for a well-polished gem! I'm a big fan of this sort of work.

arunagw:
This has been fixed by #303.

PR #268, opened 2020-12-29
Skip parsing response if content is not JSON

nickbair-nylas:
## Issue
The changes to response parsing that were released in v4.6.3 broke file downloads and any other API calls that did not return JSON.  Previously parse errors were rescued, but that was changed in v4.6.3.

## Fixes
1. Check the content type in the response headers, if the response is `application/json` parse it, in all other cases skip parsing.

I confirm that this contribution is made under the terms of the MIT license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #303, opened 2021-05-30
Fix exception raise when content-type is not JSON

arunagw:
Background:
Currently if server returns any error where content-type is not a JSON
is not raising error in `Nylas` namespace and returns
`RestClient::Response` instead. This issue is also raised in #274
and #272. It has been found that it breaks after #268 where we 
have fixed parsing error if content type is not json.

This PR makes sure that whatever content-type is returned from
API in error response it always raises error in `Nylas`
namespace.

Changes:
- Update `Nylas::HttpClient` to parse error response beforing
  raising exception.

### DIFF
--- a/lib/nylas/http_client.rb
+++ b/lib/nylas/http_client.rb
@@ -177,13 +177,13 @@ module Nylas
 
     def handle_failed_response(result:, response:)
       http_code = result.code.to_i
-      return if http_code == 200
 
       handle_anticipated_failure_mode(http_code: http_code, response: response)
       raise UnexpectedResponse, result.msg if result.is_a?(Net::HTTPClientError)
     end
 
     def handle_anticipated_failure_mode(http_code:, response:)
+      return if http_code == 200
       exception = HTTP_CODE_TO_EXCEPTIONS.fetch(http_code, APIError)
       response = parse_response(response)
 
--- a/lib/nylas/http_client.rb
+++ b/lib/nylas/http_client.rb
@@ -178,6 +178,7 @@ module Nylas
 
     def handle_anticipated_failure_mode(http_code:, response:)
       return if http_code == 200
+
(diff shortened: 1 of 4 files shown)


lib/nylas/http_client.rb
```diff
@@ -178,10 +178,15 @@ def handle_failed_response(result:, response:)
 
     def handle_anticipated_failure_mode(http_code:, response:)
       return if http_code == 200
-      return unless response.is_a?(Hash)
 
       exception = HTTP_CODE_TO_EXCEPTIONS.fetch(http_code, APIError)
-      raise exception.new(response[:type], response[:message], response.fetch(:server_error, nil))
+      response = parse_response(response)
```
adarsh, line 183: Looks like you are overwriting this variable? What do you think about using a new name?

spec/nylas/collection_spec.rb
```diff
@@ -165,4 +165,40 @@ def example_instance_hash
...
+        model = instance_double("Model")
+        allow(model).to receive(:searchable?).and_return(true)
+        allow(model).to receive(:resources_path)
+        collection = described_class.new(model: model, api: api)
+        stub_request(:get, "https://api.nylas.com/search?limit=100&offset=0&q=%7B%7D")
+          .to_return(status: code, body: {}.to_json)
+
+        expect do
+          collection.search({}).last
+        end.to raise_error(error)
```
adarsh, line 200: Nice block

adarsh (approved):
One style comment, lgtm

mrashed-dev (approved):
Looks good to me!

lib/nylas/http_client.rb
arunagw: Yes. Done here 3a3ea3a305e30f7a869daebf7affe4ba9a06207a

PR #304, opened 2021-06-04
[61539] Draft version not updating after `update`

mrashed-dev:
# Description
When making updates to the draft object on the server side (via `.update`) the version number is incremented upon successful update. However, the local object does not update the value when it's returned from the server. This causes an issue when, the user attempts to re-update the draft a second time and the version number has not been incremented. Now, when calling `.update` the version is updated with the value returned from the payload.

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 * Add `save_all_attributes` and `update_all_attributes` methods to support
   nullifying attributes. #299
 * Fix issue where files couldn't be attached to drafts #302
+* Fix exception raise when content-type is not JSON #303
+* Fix issue where draft version wouldn't update after `update` or `save` #304
 
 ### 5.0.0 / 2021-05-07
 
--- a/lib/nylas/draft.rb
+++ b/lib/nylas/draft.rb
@@ -42,10 +42,7 @@ module Nylas
       extract_file_ids!
       data[:file_ids] = file_ids
 
-      response = super
-      return unless response[:version]
-
-      self.version = response[:version]
+      super
     end
 
     def create
--- a/lib/nylas/draft.rb
+++ b/lib/nylas/draft.rb
@@ -39,6 +39,7 @@ module Nylas
     transfer :api, to: %i[events files folder labels]
 
     def update(**data)
+      self.files = data[:files] if data[:files]
       extract_file_ids!
       data[:file_ids] = file_ids
 
--- a/lib/nylas/draft.rb
+++ b/lib/nylas/draft.rb
@@ -42,7 +42,10 @@ module Nylas
       extract_file_ids!
       data[:file_ids] = file_ids
 
-      super
+      response = super
+      return unless response[:version]
+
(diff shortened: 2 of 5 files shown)


lib/nylas/model.rb
```diff
@@ -64,7 +64,6 @@ def update(**data)
         path: resource_path,
         query: query_params
       )
-      true
```
arunagw, line 67: Not sure if it's a good idea to change response here as existing apps might be check if `update` returns `true` or not

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 * Add `save_all_attributes` and `update_all_attributes` methods to support
   nullifying attributes. #299
 * Fix issue where files couldn't be attached to drafts #302
+* Fix exception raise when content-type is not JSON #303
```
arunagw, line 7: Thanks for adding missing CHANGELOG

mrashed-dev: No problem!

lib/nylas/model.rb
mrashed-dev: Yeah I know that this may be an issue for pre-existing apps, but without making this change, then we won't be able to know the version number in the `draft` class:

```
response = super
return unless response[:version]

self.version = response[:version]
```

`response` would just be `true` and not a JSON. The alternative is that we check if it's `true` and increment the version number ourselves. But I'm not a fan of making the SDK 'predict' an attribute locally in case the Nylas API changes in the future. Let me know what you think.

lib/nylas/model.rb
```diff
@@ -58,12 +58,13 @@ def update(**data)
...
 
       attributes.merge(**data)
-      execute(
+      result = execute(
         method: :put,
         payload: attributes.serialize_for_api(keys: data.keys),
         path: resource_path,
         query: query_params
       )
+      attributes.merge(result) if result
```
mrashed-dev, line 67: Very nice!

lib/nylas/model.rb
mrashed-dev: Agreed, I like your approach much better.

lib/nylas/model.rb
```diff
@@ -134,6 +128,16 @@ def save_call
       )
     end
 
+    def update_call(payload)
```
adarsh, line 131: Nice extraction!

lib/nylas/draft.rb
```diff
@@ -39,6 +39,7 @@ class Draft
     transfer :api, to: %i[events files folder labels]
 
     def update(**data)
+      self.files = data[:files] if data[:files]
```
adarsh, line 42: Be sure to test both forks of conditionals when you introduce them into code.

adarsh (commented):
Needs another test or two but otherwise very nice work!

lib/nylas/draft.rb
mrashed-dev: We had test cases but they've been further clarified to test for this specific use case.

PR #307, opened 2021-06-16
[61632] Add `Room Resource` support

mrashed-dev:
# Description
This adds support for the `/resource` endpoint and the Nylas `Room Resource` object. More information can be found [in the docs](https://developer.nylas.com/docs/api/#tag--Room-Resources).

# Usage
Currently, the `/resource` endpoint only supports the `GET` operation without any extra functionality like filtering. To get a list of all room resource objects attached to your account,  you can do the following:

```
#!/usr/bin/env ruby
require 'nylas'

# Initialize and connect to the Nylas client
nylas = Nylas::API.new(
  app_id: CLIENT_ID,
  app_secret: CLIENT_SECRET,
  access_token: ACCESS_TOKEN
)

resources = nylas.room_resources
```

# License

### DIFF
--- a/lib/nylas/api.rb
+++ b/lib/nylas/api.rb
@@ -123,6 +123,11 @@ module Nylas
       @messages ||= Collection.new(model: Message, api: self)
     end
 
+    # @return[Collection<RoomResource>] A queryable collection of {RoomResource} objects
+    def room_resources
+      @room_resources ||= Collection.new(model: RoomResource, api: self)
+    end
+
     # Revokes access to the Nylas API for the given access token
     # @return [Boolean]
     def revoke(access_token)
--- a/lib/nylas/room_resource.rb
+++ b/lib/nylas/room_resource.rb
@@ -6,7 +6,7 @@ module Nylas
   class RoomResource
     include Model
     self.resources_path = "/resources"
-    allows_operations(listable: true, showable: true)
+    allows_operations(listable: true)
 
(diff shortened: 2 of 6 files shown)


arunagw:
@mrashed-dev great work! I think we also need to mention that we need add one more scope to fetch this information something like `room_resources.read_only` based on the documentation.

lib/nylas/api.rb
```diff
@@ -123,6 +123,11 @@ def messages
       @messages ||= Collection.new(model: Message, api: self)
     end
 
+    # @return[Collection<RoomResource>] A queryable collection of {RoomResource} objects
+    def room_resources
```
arunagw, line 127: I think we need spec for this one to make sure we call proper API endpoint.

lib/nylas/room_resource.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Nylas
+  # Ruby representation of a Nylas Room Resource object
+  # @see https://developer.nylas.com/docs/api/#tag--Room-Resources
+  class RoomResource
+    include Model
+    self.resources_path = "/resources"
+    allows_operations(listable: true, showable: true)
```
arunagw, line 9: Can it be `showable`? I don't see any route to fetch a particular room. I might be looking at wrong place.

dtom90 (approved):
LGTM 👍

PR #309, opened 2021-07-07
cleanup: unused service_domain removed

olegykz:
During working with Ireland region we found out that we need to use Ireland API server URL in `api_server` argument for API client instantiation. 
Additionally to this parameter, we met `service_domain` argument which seems to be not used anymore. 

The last usage of this parameter was found at https://rdoc.info/gems/inbox/1.3.0/Inbox%2FAPI:url_for_authentication - in order to avoid any further confusion this PR is intended to remove unused argument.


mrashed-dev (approved):
Looks good to me, thanks for identifying this and cleaning it up!

PR #319, opened 2021-08-04
Add messages attribute to Thread.

GuYifan:
This will help deserialize messages for GET requests sent to /threads/{{id}}?view=expanded.

Issue is that when tries to get thread with view=expanded, it doesn't deserialize messages.


mrashed-dev:
Great find and thank you for your contribution 🎉 Also much appreciate the added test case! I've just added a changelog entry and merging it in, it will be included in the next Ruby release!

GuYifan:
> Great find and thank you for your contribution 🎉 Also much appreciate the added test case! I've just added a changelog entry and merging it in, it will be included in the next Ruby release!

Sounds good! Cheers!

PR #323, opened 2021-08-25
[67394] Fix updating `Draft` failing if no version number provided

mrashed-dev:
# Description
The following is the code snippet is found on our Nylas docs for Ruby PUT `/drafts/{id}`: https://developer.nylas.com/docs/api/#put/drafts/id

```ruby
draft_to_change = api.drafts.find("51si***")
draft_to_change.update(display_name: "My Renamed Draft")
```

However in practice this fails. The reason is because the API expects a version number to be passed in with a PUT call. In the Ruby SDK we only pass the updated parameters to the API, so if the user doesn't explicitly send in the `version` number, the SDK doesn't do it for them and triggers the error. To adhere to the documentation and the API's expectation, we should send the version number in the payload if the user never adds it in explicitly.

# License

### DIFF
--- a/lib/nylas/draft.rb
+++ b/lib/nylas/draft.rb
@@ -39,9 +39,12 @@ module Nylas
     transfer :api, to: %i[events files folder labels]
 
     def update(**data)
+      # If files are provided, replace files with file IDs
       self.files = data[:files] if data[:files]
       extract_file_ids!
       data[:file_ids] = file_ids
+      # If version is not provided, add version or request will fail
+      data[:version] = version unless data[:version]
 
       super
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### Unreleased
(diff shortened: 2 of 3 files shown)


lib/nylas/draft.rb
```diff
@@ -39,9 +39,12 @@ class Draft
     transfer :api, to: %i[events files folder labels]
 
     def update(**data)
+      # If files are provided, replace files with file IDs
       self.files = data[:files] if data[:files]
       extract_file_ids!
       data[:file_ids] = file_ids
+      # If version is not provided, add version or request will fail
+      data[:version] = version unless data[:version]
```
dtom90, line 47: Could we add a test case for this in `draft_spec.rb`?

dtom90 (approved):
Confirmed that this is the expected behavior.

As a TODO, might be good to add some test cases for this

lib/nylas/draft.rb
mrashed-dev: Done, thank you for pointing that out!

PR #324, opened 2021-08-30
Change drafts example so that it works even when a user has no existing drafts

dtom90:
Move "Creating a draft" to the top so that subsequent steps (e.g. Retrieving a particular draft) will still work even if a user has no pre-existing drafts


mrashed-dev (approved):
Looks good to me!

PR #325, opened 2021-09-16
[68888] Modify Draft Sending logic

mrashed-dev:
# Description
We shouldn't be updating a draft every time the user tries to to send a draft. This causes an unnecessary PUT call, and potentially causes a delay in the syncing process for the end user. So now, calling `.send` on a draft will not trigger an update/PUT call. Also to make up for this change, if a user wants to send a draft that they haven't saved (have not triggered a POST `/drafts` call), the SDK will send the full body in the POST call to `/send` allowing the user to send a message directly without having to create and save a draft object first.

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Unreleased
 * Fix update draft failing if version is not explicitly set
+* Fix draft `.send` logic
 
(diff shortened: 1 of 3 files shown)


dtom90 (approved):
LGTM 👍

PR #327, opened 2021-09-27
Update draft send method to send existing draft when tracking is enabled

iansmith9876:
## Description
This fixes a bug where if tracking information is enabled on a draft it sends directly rather than sending the existing draft. If a draft has already been saved this means the actual draft message never gets sent and a new message is created. The message should instead only be sent directly if it hasn't been saved and otherwise conditionally add the tracking information to the draft message.

I've also updated the draft spec to reflect this change. Previously it was expecting:
`{\"id\":\"draft-1234\",\"version\":5,\"tracking\":{\"links\":true,\"opens\":true,\"thread_replies\":true,\"payload\":\"this is a payload\"}}`

It should instead expect `draft_id` rather than `id`:
`{\"draft_id\":\"draft-1234\",\"version\":5,\"tracking\":{\"links\":true,\"opens\":true,\"thread_replies\":true,\"payload\":\"this is a payload\"}}`

## License

### DIFF
--- a/lib/nylas/draft.rb
+++ b/lib/nylas/draft.rb
@@ -56,9 +56,11 @@ module Nylas
     end
 
     def send!
-      return execute(method: :post, path: "/send", payload: to_json) if tracking || !id
+      return execute(method: :post, path: "/send", payload: to_json) unless id
 
-      execute(method: :post, path: "/send", payload: JSON.dump(draft_id: id, version: version))
+      data = { draft_id: id, version: version }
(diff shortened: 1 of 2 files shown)


philrenaud:
Hi @iansmith9876 ! Thanks so much for putting this together. We'll make sure everything looks good shortly and prepare a release if so.

dtom90 (approved):
LGTM! 👍

iansmith9876:
Thanks for the quick review and release! 👏

PR #340, opened 2021-12-07
[75450] Fix bug where updating an Event results in an API error

mrashed-dev:
# Description
A recent API update caused updating an event using the SDK to return an error because the SDK sends a Participant's status as well as an object key within the Event.when object. This PR also modifies how we serialize child objects for API calls, and ensure if we don't want to include read_only values that we enforce it throughout the child objects.

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### Unreleased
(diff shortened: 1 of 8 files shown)


dtom90 (approved):
LGTM 👍

PR #362, opened 2022-04-27
Update app.rb to replace deprecated Google scopes

maxw-nylas:
# Description
https://www.google.com/m8/feeds/ is a deprecated scope. We have users relying on this example code for our authentication. This code won't work unless the scope mentioned is replaced with https://www.googleapis.com/auth/contacts

Related CC PR here (not public): https://github.com/nylas/cloud-core/pull/5960/files

# License


mrashed-dev (approved):
Looks good to me, good catch @maxw-nylas!

PR #367, opened 2022-05-10
Fix account not filterable error

nickbair-nylas:
# Description
Fix account not filterable error when querying for accounts with metadata.  The following now works:

```
nylas.accounts.where(metadata_pair: { "provider": "eas" })
```

# License


mrashed-dev (approved):
Looks good to me!

PR #375, opened 2022-07-26
Fix calendar availability failing when `round_robin` is `nil`

mrashed-dev:
# Description
This PR address the issue where, if `round_robin` is left as `nil`, it always yields an error. This is because the Nylas Docs were out of date for the availability endpoint. `round_robin` is optional, and when omitted, returns collective availability. However, sending `null` is not valid despite what it says in the documentation. Now, we omit the value from the json payload if the value is `nil`.

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### Unreleased
+* Add missing hosted authentication parameters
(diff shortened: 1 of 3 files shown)


quzhi1:
CI is failing

mrashed-dev:
Thanks @quzhi1, CI is fixed now :)

quzhi1 (approved):

PR #379, opened 2022-08-08
fix Collection new method to expand keyword args

icheishvili:
# Description
ruby 3 does not automatically expand keyword arguments
the create method in this class already does double-splat expansion
this change makes the new method match and fixes a ruby 3 regression

# License


mrashed-dev (approved):
Good catch, thanks for your contribution @icheishvili!!! 🚀

PR #384, opened 2022-09-23
Add timezone fields to the When class

mrashed-dev:
# Description
This PR adds the `timezone`, `start_timezone` and `end_timezone` fields. This SDK also now uses tzinfo as a dev dependency to validate the strings set by the user are indeed IANA database names.

# License


quzhi1:
Failing CI

quzhi1 (approved):

PR #385, opened 2022-10-11
Adds visibility attribute in Event.rb

brunossilveira:
This PR adds the `visibility` attribute to the `Event` class.

### DIFF
--- a/lib/nylas/event.rb
+++ b/lib/nylas/event.rb
@@ -43,6 +43,7 @@ module Nylas
     attribute :reminder_minutes, :string
     attribute :reminder_method, :string
     attribute :job_status_id, :string, read_only: true
+    attribute :visibility, :string, read_only: true
 
     attr_accessor :notify_participants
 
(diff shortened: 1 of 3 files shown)


lib/nylas/event.rb
```diff
@@ -43,6 +43,7 @@ class Event
     attribute :reminder_minutes, :string
     attribute :reminder_method, :string
     attribute :job_status_id, :string, read_only: true
+    attribute :visibility, :string, read_only: true
```
mrashed-dev, line 46: Visibility is actually writeable for Google events (which is the only provider that uses this field).

```suggestion
    attribute :visibility, :string
```

mrashed-dev (changes requested):
@brunossilveira Thank you so much for opening this PR! It looks good mostly, just made a comment about the readability of the variable. Once amended I'll approve and merge. 🚀

brunossilveira:
Done!

PR #386, opened 2022-10-20
Add support for calendar colors

mrashed-dev:
# Description
This PR adds support for the new colors field in the Calendar model. Please note that this field is read only, and for Microsoft calendars only.

# License

### DIFF
--- a/lib/nylas/calendar.rb
+++ b/lib/nylas/calendar.rb
@@ -40,7 +40,6 @@ module Nylas
     end
 
     def events
-      event = api.events
       api.events.where(calendar_id: id)
     end
   end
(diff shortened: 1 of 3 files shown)


lib/nylas/calendar.rb
```diff
@@ -29,6 +29,7 @@ class Calendar
     attribute :read_only, :boolean
     attribute :metadata, :hash
     attribute :job_status_id, :string, read_only: true
+    attribute :color, :integer, read_only: true
```
quzhi1, line 32: Are you sure calendar colors are integer, not string? For passthru, color is defined as a string: https://passthru.pages.dev/#tag/Calendar/Calendar-Object

mrashed-dev: @quzhi1 good catch.. I think I was basing the implementation on an earlier commit where we directly reflected the EWS implementation (https://learn.microsoft.com/en-us/openspecs/exchange_server_protocols/ms-oxprops/1fdd4d65-8c85-4ea3-95c4-5d5f55e33eb8).

I'll amend this PR and fix it, thanks!

quzhi1 (approved):

PR #390, opened 2022-11-15
Add support for async send using the sendnow query parameter

rossnelson:
# Description

Add support for async message sending using the sendnow query parameter

https://developer.nylas.com/docs/connectivity/email/outbox-guide/#sending-messages

# License
I confirm that this contribution is made under the terms of the MIT license and
that I have the authority necessary to make this contribution on behalf of its
copyright owner.


rossnelson:
Well, nvm. I just noticed the previous async send is deprecated and there is soon to be a v2 send. https://developer.nylas.com/docs/connectivity/email/v2_send/

PR #395, opened 2022-12-15
Issue-394: Add disable_provider_selection kwarg to Api.authentication_url

PickertJoe:
# Description
We need the option to disable the "Select different provider" button in Nylas' hosted auth when we are trying to authenticate against a specific provider. This is necessary to prevent our users from escaping the intended auth flow and selecting a provider that we do not support.

This commit addresses this need by adding and optional keyword argument to the Api.authentication_url method and, if it is provided, including it in the params payload appended on the call to the /oauth/authorize endpoint. Tests have been added to account for this change and I tested it locally to confirm that the "Select different provider" button was no longer visible when I included disable_provider_selection: true in my call to the authentication_url method.

# License

### DIFF
--- a/lib/nylas/api.rb
+++ b/lib/nylas/api.rb
@@ -36,7 +36,7 @@ module Nylas
     end
 
     def authentication_url(redirect_uri:, scopes:, response_type: "code", login_hint: nil, state: nil,
-                           provider: nil, redirect_on_error: nil)
+                           provider: nil, redirect_on_error: nil, disable_provider_selection: nil)
       params = {
         client_id: app_id,
         redirect_uri: redirect_uri,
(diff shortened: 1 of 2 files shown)


PickertJoe:
Hey @mrashed-dev @andreccosta! I'm not quite sure who to tag as reviewers for this PR, but I saw you'd contributed recently and was curious if you'd mind taking a look at this for me. I'd be happy to share more about these changes and our need for them if needed. Thanks so much!

kellyredding (approved):
@PickertJoe Can I approve this? We'll see! 🚀 

I think this change looks great! Nice that it was so easy to make.

mrashed-dev (approved):
Thanks for your contribution @PickertJoe, looks good to me!

Issue #407, opened 2023-03-10
Event (and maybe other objects) missing 'updated_at' attribute.


tvongaza:
**Describe the bug**
The web API returns an `updated_at` field, however it is not included in the ruby library as a defined attribute.

**To Reproduce**
Some steps involved to reproduce the bug and any code samples you can share.
```
- Call the events api via curl and see the event having an updated_at field.
- Call the same event via the ruby sdk and see the ruby object does not have an updated_at field.
```

**Expected behavior**
The ruby events object should have the same fields, such as `updated_at` as the web api.

**SDK Version:**
Latest as of March 9, 2023.

**Additional context**
I monkey patched the event class as follows, which gets me the updated_at field as expected.

```
Nylas::Event.class_eval do
  attribute :updated_at, :integer, read_only: true
end
```

PR #410, opened 2023-03-18
Added event.updated_at

tvongaza:
# Description
The API returns `updated_at` for events, but it is not included in the ruby client event model. This PR adds the `updated_at` attribute and related spec. This was referenced in #407 which was closed by #409, but it didn't actually add the `updated_at` attribute.

# License


mrashed-dev (approved):
Thanks for another contribution, @tvongaza !

PR #452, opened 2024-01-05
Fix incorrect PKCE code challenge generation

mrashed-dev:
# Description
This PR fixes the PKCE code challenge generation; the correct method the API wants is for us to base64 encode the hex string as opposed to base64 encoding resulting hashed bytearray directly.

# License


AaronDDM (approved):

atejada (approved):
All good 👍

PR #453, opened 2024-01-08
Fixed incompatible types when building OAuth2 URL

mrashed-dev:
# Description
We were encoding our URL prematurely, causing an error to be thrown when setting additional parameters (PKCE, etc.)

# License


atejada (approved):
All good to 🚀

PR #456, opened 2024-01-23
support draft filtering

rfgil:
# Description
Support filtering draft messages, using the `/drafts` endpoint according to the [documentation](https://developer.nylas.com/docs/api/v2/#get-/drafts). 

As an example, this allows querying for drafts on a given a `thread_id` using:
```
nylas_api.drafts.where(thread_id:)
```

# License


mrashed-dev (approved):
Thank you @rfgil! 🚀

PR #457, opened 2024-01-24
Update Client.rb to include SmartCompose

atejada:
# Description

SmartCompose is missing from Client.rb

https://nylas.atlassian.net/browse/TW-2501

# License


atejada:
My bad...didn't realized that SmartCompose was actually inside Messages 🤦🏻‍♂️

PR #466, opened 2024-02-16
Move Detect Provider from Auth to Connectors

atejada:
# Description
Currently Detect Provider is on Auth, but it should be on Connectors too

https://nylas.atlassian.net/browse/TW-2599

# License


mrashed-dev:
No longer needed as we are not moving detect provider.

PR #474, opened 2024-04-05
Support custom headers

axfcampos:
# Description

Support sending a message with customer headers, on the `/send` endpoint, according to the documentation: https://developer.nylas.com/docs/api/v2/#post-/send

# License


mrashed-dev (approved):
LGTM, thanks @axfcampos!

PR #483, opened 2024-08-19
Fix sending attachments less than 3MB

mrashed-dev:
# Description
We were not encoding the attachments properly for attachments less than 3MB. They should be b64 encoded. Now it should handle it correctly.

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### Unreleased
+* Fixed sending attachments less than 3MB
+
 ### 6.1.0 / 2024-06-25
 * Added query support for messages
 * Added support for clean messages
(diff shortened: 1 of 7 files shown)


kraju3:
@mrashed-dev Thanks for raising this PR. I have an edge case:

What if the clients doesn't specify a file but instead specify the base64 encoded content (i.e If they are not using our utility function)

atejada (approved):

atejada (approved):
All good 👍

mrashed-dev:
@kraju3 function comments updated along with code to support this operation:

```ruby
# @param request_body [Hash] The values to create the message with.
    #   If you're attaching files, you must pass an array of [File] objects, or
    #   you can pass in base64 encoded strings if the total attachment size is less than 3mb.
    #   You can also use {FileUtils::attach_file_request_builder} to build each object attach.
```

PR #487, opened 2024-09-12
Enable SDK to reattach large files to messages on retry

mrashed-dev:
# Description
This PR enables the SDK to re-attach large files when trying to retry the same request body. If a `file_path` is provided with the attachment request and the file stream is closed, the SDK will attempt to reopen it. The helper function `attach_file_request_builder` will now set the `file_path` in the object, so if you're currently using that helper function, the attachment will be ready for reattachment. 

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### Unreleased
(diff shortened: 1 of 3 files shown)


atejada (approved):
LGTM 👍

PR #490, opened 2024-09-17
Loosen rest-client version

lekemula:
# Description

Hi there,

In my current organisation, we are trying to upgrade to nylas API v3, however nylas 6.x Ruby SDK conflicts with one of our [chargebee-ruby ](https://github.com/chargebee/chargebee-ruby) sub-dependency:

```
chargebee (2.43.0)
      rest-client (>= 1.8, <= 2.0.2)
nylas (6.1.1)
  	  rest-client (~> 2.1, < 3.0) 
```

While there's already a PR to upgrade chargebee usage of rest-client, https://github.com/chargebee/chargebee-ruby/pull/83, we're unsure whether that will land on time, so we were wondering whether for the time being we could loosen up the version on your side?

When running the CI against `rest-client 2.0.0` there seems to be ✅. Do you see any repercussions against doing that?

```
❯ bi
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Using rest-client 2.0.0 (was 2.1.0)
Bundle complete! 12 Gemfile dependencies, 47 gems now installed.
Use `bundle info [gemname]` to see where a bundled gem is installed.
❯ be rspec
....................................................................................................................................................................

Finished in 0.05607 seconds (files took 0.39409 seconds to load)
164 examples, 0 failures
```

This is currently blocking us from migrating to Nylas API v3.

Thanks in advance for your help.

# License

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,4 +2,5 @@
 
 source "https://rubygems.org"
 
+gem "rest-client", "2.0.0"
 gemspec name: "nylas"
--- a/Gemfile
+++ b/Gemfile
@@ -2,5 +2,4 @@
 
 source "https://rubygems.org"
 
-gem "rest-client", "2.0.0"
 gemspec name: "nylas"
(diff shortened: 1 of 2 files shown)


mrashed-dev (approved):
@lekemula Looking at the changelog from 2.0 -> 2.1 the changes look like they are not critical to how the SDK uses them and as long as the CIs passed then we should be good! Thanks and will have this in the next release.

PR #492, opened 2024-09-24
Added query support for folders

### LINKED ISSUE #488: Add `query_params` to the `folders.list(..)` method
There is no way to retrieve a second page of folders since the [list method](https://github.com/nylas/nylas-ruby/blob/main/lib/nylas/resources/folders.rb#L18) in Folders only takes the identifier parameter.

If the query_params argument is added (for example https://github.com/nylas/nylas-ruby/blob/main/lib/nylas/resources/messages.rb#L30 we'd be able to pass in a cursor for pagination.

We have worked around it by building the api call ourself but would help clean things up if this was made to be consistent with the other list methods.

I do see the cursor is already being returned, so only allowing it to be passed in via query_params should solve the issue.

mrashed-dev:
# Description
This PR adds the ability to pass query parameters for listing folders. Closes #488.

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ### Unreleased
+* Added query support for folders
 * Enable SDK to reattach large files to messages on retry
 
 ### 6.1.1 / 2024-08-20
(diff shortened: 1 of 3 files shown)


atejada (approved):
All good 👍🏻

PR #496, opened 2024-10-16
Allow custom filename when adding an attachment

loomchild:
# Description
When adding an attachment to an email, the filename that will be sent is determined by the provided file path. This is problematic when using `Tempfile` because part of the path will be random. I propose accepting an optional `filename` parameter in `FileUtils.attach_file_request_builder`.

# License

### DIFF
--- a/lib/nylas/utils/file_utils.rb
+++ b/lib/nylas/utils/file_utils.rb
@@ -92,8 +92,8 @@ module Nylas
     # Build the request to attach a file to a message/draft object.
     # @param file_path [String] The path to the file to attach.
     # @return [Hash] The request that will attach the file to the message/draft
-    def self.attach_file_request_builder(file_path)
-      filename = File.basename(file_path)
+    def self.attach_file_request_builder(file_path, filename = nil)
+      filename ||= File.basename(file_path)
       content_type = MIME::Types.type_for(file_path)
       content_type = if !content_type.nil? && !content_type.empty?
                        content_type.first.to_s
--- a/lib/nylas/utils/file_utils.rb
+++ b/lib/nylas/utils/file_utils.rb
@@ -91,6 +91,7 @@ module Nylas
 
     # Build the request to attach a file to a message/draft object.
     # @param file_path [String] The path to the file to attach.
+    # @param filename [String] The name of the attached file. Optional, derived from file_path by default.
     # @return [Hash] The request that will attach the file to the message/draft
     def self.attach_file_request_builder(file_path, filename = nil)
       filename ||= File.basename(file_path)
--- a/spec/nylas/utils/file_utils_spec.rb
+++ b/spec/nylas/utils/file_utils_spec.rb
@@ -42,6 +42,21 @@ describe Nylas::FileUtils do
         file_path: file_path
       )
     end
+
+    it "accepts optional filename parameter" do
+      file_path = "/path/to/file.txt"
+      filename = "customm-file.txt"
+
+      attach_file_req = described_class.attach_file_request_builder(file_path, filename)
+
+      expect(attach_file_req).to eq(
+        filename: "customm-file.txt",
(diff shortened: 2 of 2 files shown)


lib/nylas/utils/file_utils.rb
```diff
@@ -92,8 +92,8 @@ def self.handle_message_payload(request_body)
     # Build the request to attach a file to a message/draft object.
     # @param file_path [String] The path to the file to attach.
     # @return [Hash] The request that will attach the file to the message/draft
-    def self.attach_file_request_builder(file_path)
-      filename = File.basename(file_path)
+    def self.attach_file_request_builder(file_path, filename = nil)
```
loomchild, line 95: Alternatively, it could also be named parameter: `filename: nil`.

loomchild:
@SubashPradhan could you please tell me what do you think about this PR, is there a chance of merging it?

SubashPradhan:
> @SubashPradhan could you please tell me what do you think about this PR, is there a chance of merging it?

@loomchild Thanks for looking into this, while sending an attachment, you can configure the custom `file_name` on the `attachments` field itself.

Have you tried the above approach, example of custom `filename` would look as below, in the below example the filename will be reflected as `different_name.jpeg` in the email.

```
 file_path = "/path/file_name.jpeg"
  base64_content = file_to_base64(file_path)
  buffer = Base64.decode64(base64_content)
  stream = StringIO.new(buffer)

  request_body = {
    subject: "Hello",
    to: [{ name: "My Friend", email: "test.user@gmail.com" }],
    body: "Hello, World!",
    attachments: [{
      content: stream,
      content_id: 'image_1_Xf0apsUfZ',
      filename: 'different_name.jpeg',
      size: 370,
      content_type: 'image/jpeg'
    }]
  }

Can you tell me if this is something you were aiming for? Thanks

**Comment by loomchild** [comment]
No. I don't want to load the entire attachment to memory. It also won't work for files > 3M, since they require special handling on Nylas and simple Base64 encoding won't work.

That's why I want to use `attach_file_request_builder` helper provided by Nylas. Obviously, I can avoid using it altogether and just copy-paste your code to our app, or link a file to a different name (that's what I am currently doing). But I am trying to avoid that by suggesting an improvement.

File: lib/nylas/utils/file_utils.rb [hunk 2]
```diff
@@ -92,8 +92,8 @@ def self.handle_message_payload(request_body)
     # Build the request to attach a file to a message/draft object.
     # @param file_path [String] The path to the file to attach.
```
SubashPradhan, line 93: Can we add doc comment?

SubashPradhan:
> No. I don't want to load the entire attachment to memory. It also won't work for files > 3M per your documentation, since they require special multi-part request to Nylas API and simple Base64 embedded in JSON won't do.
> 
> That's why I want to use `attach_file_request_builder` helper provided by Nylas. Obviously, I can avoid using it altogether and just copy-paste your code to our app, or link a file to a different name to make the helper happy (that's what I am currently doing). But I am trying to avoid that hassle and use standard Tempfile.
> 
> That's why I am suggesting a simple improvement in your SDK, but if you are not interested, feel free to drop it.

I see, thanks for the explanation, we should be good to merge the doc comment is added.

lib/nylas/utils/file_utils.rb
loomchild: Sure.

SubashPradhan (approved):
Thanks for contributing 🎉

loomchild:
Sure, thanks for approving :slightly_smiling_face:. Looking forward to the next release.

loomchild:
Unfortunately it doesn't work for attachments > 3MB (unfortunately I didn't test it earlier).
I created a follow-up: https://github.com/nylas/nylas-ruby/pull/502

Issue #499, opened 2024-11-01
Fetching a folder should support select field


bmulholland:
https://developer.nylas.com/docs/api/v3/ecc/#get-/v3/grants/-grant_id-/folders/-folder_id- has a select option

https://github.com/nylas/nylas-ruby/blob/main/lib/nylas/resources/folders.rb#L31 does not

bmulholland:
(Hi @AaronDDM! Hope you're well :) )

SubashPradhan:
Closing this issue, the fix has been released in v6.2.1.

PR #500, opened 2024-11-04
Add select field to folders find method

SubashPradhan:
# Description
- Add `select` field to `GET folders/id` 
- https://github.com/nylas/nylas-ruby/issues/499

# License

### DIFF
--- a/lib/nylas/resources/folders.rb
+++ b/lib/nylas/resources/folders.rb
@@ -27,6 +27,7 @@ module Nylas
     #
     # @param identifier [String] Grant ID or email account to query.
     # @param folder_id [String] The id of the folder to return.
+    # @param query_params [Hash, nil] Query params to pass to the request.
     # @return [Array(Hash, String)] The folder and API request ID.
     def find(identifier:, folder_id:, query_params: nil)
       get(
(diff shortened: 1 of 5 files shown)


lib/nylas/resources/folders.rb
```diff
@@ -28,9 +28,10 @@ def list(identifier:, query_params: nil)
     # @param identifier [String] Grant ID or email account to query.
     # @param folder_id [String] The id of the folder to return.
```
AaronDDM, line 29: Ulster the doc comments

spec/nylas/resources/folders_spec.rb
```diff
@@ -42,7 +42,7 @@
       folder_id = "5d3qmne77v32r8l4phyuksl2x"
       path = "#{api_uri}/v3/grants/#{identifier}/folders/#{folder_id}"
       allow(folders).to receive(:get)
-        .with(path: path)
+        .with(path: path, query_params: nil)
```
AaronDDM, line 45: Add a new test that tests the query param as well

bmulholland:
Thanks for this! Just noticed that threads.find is also missing the same. Any chance we could get that while you're at it?

SubashPradhan:
> Thanks for this! Just noticed that threads.find is also missing the same. Any chance we could get that while you're at it?

Sure, thanks for pointing that out.

AaronDDM (approved):

bmulholland:
FYI I've tried this out in production and it's working well. Thank you!

PR #502, opened 2024-11-13
Support custom attachment filename for large files

loomchild:
# Description
Add support for attachments > 3MB.
Follow-up of #496.

# License

### DIFF
--- a/lib/nylas/utils/file_utils.rb
+++ b/lib/nylas/utils/file_utils.rb
@@ -36,6 +36,13 @@ module Nylas
           file = File.open(attachment[:file_path], "rb")
         end
 
+        # Setting original filename and content type if available. See code rest-client#lib/restclient/payload.rb
+        filename = attachment[:filename] || attachment["filename"]
+        file.define_singleton_method(:original_filename) { filename } if filename
+
+        content_type = attachment[:content_type] || attachment["content_type"]
+        file.define_singleton_method(:content_type) { content_type } if content_type
+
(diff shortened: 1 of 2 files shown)


lib/nylas/utils/file_utils.rb
```diff
@@ -36,6 +36,13 @@ def self.build_form_request(request_body)
           file = File.open(attachment[:file_path], "rb")
         end
 
+        # Setting original filename and content type if available. See code rest-client#lib/restclient/payload.rb
+        filename = attachment[:filename] || attachment["filename"]
+        file.define_singleton_method(:original_filename) { filename } if filename
+
+        content_type = attachment[:content_type] || attachment["content_type"]
+        file.define_singleton_method(:content_type) { content_type } if content_type
```
loomchild, line 44: This might not be required - added as a bonus.

lib/nylas/utils/file_utils.rb
```diff
@@ -36,6 +36,13 @@ def self.build_form_request(request_body)
           file = File.open(attachment[:file_path], "rb")
         end
 
+        # Setting original filename and content type if available. See code rest-client#lib/restclient/payload.rb
```
loomchild, line 39: Using not fully documented rest-client feature to set the filename in the payload. See https://github.com/rest-client/rest-client/blob/master/lib/restclient/payload.rb#L185

SubashPradhan (approved):

PR #509, opened 2025-01-10
Handle response objects of instance String for API errors

kraju3:
# Description
Due to unexpected errors such as Bad Gateway, the response object is of Instance string, while trying to fetch the provider_error we return a Ruby error of 
`undefined method `fetch' for "Bad Gateway":String`

This PR addresses the type before parsing the provider_error property

# License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### Unreleased
+* Fixed issue where errors were not properly thrown due to an instance type of `String` instead of `Hash`
+
 ### 6.2.2 / 2024-12-02
 * Added support for private Scheduling configuration.
 * Added ability to add optional `content_id` to support inline image on `send`.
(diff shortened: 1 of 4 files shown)


nickbair-nylas (approved):
LGTM, worth adding a test case for this?

AaronDDM (commented):
Thanks Kiran! Could you add a test, ideally for a legitimate error that we're experiencing.

kraju3:
@AaronDDM @nickbair-nylas I think there are some tests regarding this in http_client.spec file? Would you be able to sanity check if that is doing the same?

nickbair-nylas:
@kraju3 the only test I see for `#throw_error` uses an error object ([here](https://github.com/nylas/nylas-ruby/blob/main/spec/nylas/handler/http_client_spec.rb#L235)).  I think you could just copy this test and pass error as a string.

kraju3:
Added a test for this change.

One of the github action is failing. I am not sure if it's related to this:
https://github.com/actions/runner-images/issues/10636

AaronDDM (approved):

PR #513, opened 2025-02-20
Add events import function to the RubySDK

kraju3:
# Description
Adding event import beta endpoint to the Events Class

# License

### DIFF
--- a/lib/nylas/resources/events.rb
+++ b/lib/nylas/resources/events.rb
@@ -94,5 +94,18 @@ module Nylas
         request_body: request_body
       )
     end
+
+    # Returns a list of recurring events, recurring event exceptions, and single events
+    # from the specifiedcalendar within a given time frame.
+    #
+    # @param identifier [String] Grant ID or email account to import events from.
+    # @param query_params [Hash] The query parameters to include in the request
+    # @return [(Array(Hash), String, String)] The list of events, API Request ID, and next cursor.
+    def events_import(identifier:, query_params:)
+      get_list(
+        path: "#{api_uri}/v3/grants/#{identifier}/events/import",
+        query_params: query_params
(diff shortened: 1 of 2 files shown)


lib/nylas/resources/events.rb
```diff
@@ -94,5 +94,18 @@ def send_rsvp(identifier:, event_id:, request_body:, query_params:)
...
       )
     end
+
+    # Returns a list of recurring events, recurring event exceptions, and single events
+    # from the specifiedcalendar within a given time frame.
+    #
+    # @param identifier [String] Grant ID or email account to import events from.
+    # @param query_params [Hash] The query parameters to include in the request
+    # @return [(Array(Hash), String, String)] The list of events, API Request ID, and next cursor.
+    def events_import(identifier:, query_params:)
```
AaronDDM, line 104:
```suggestion
    def list_event_imports(identifier:, query_params:)
```

I think we need to keep this more consistent with the rest of our APIs. Closest one I could find was scheduled send messages.

lib/nylas/resources/events.rb
```diff
@@ -94,5 +94,18 @@ def send_rsvp(identifier:, event_id:, request_body:, query_params:)
         request_body: request_body
       )
     end
+
+    # Returns a list of recurring events, recurring event exceptions, and single events
+    # from the specifiedcalendar within a given time frame.
```
AaronDDM, line 99:
```suggestion
    # Returns a list of recurring events, recurring event exceptions, and single events
    # from the specified calendar within a given time frame. This is useful when you 
    # want to import, store, and synchronize events from the time frame to your application 
```

Because the endpoint is such a strange name, I want to provide context as to why it is called import.

PR #514, opened 2025-02-20
Adding response headers to Ruby SDK

kraju3:
# Description
Adding response headers to the response Hash

# License

### DIFF
--- a/lib/nylas/handler/api_operations.rb
+++ b/lib/nylas/handler/api_operations.rb
@@ -26,7 +26,8 @@ module Nylas
       #
       # @param path [String] Destination path for the call.
       # @param query_params [Hash, {}] Query params to pass to the call.
-      # @return [Array(Array(Hash), String, String, Hash)] Nylas data array, API Request ID, next cursor, and response headers.
+      # @return [Array<Array<Hash>, String, String, Hash>]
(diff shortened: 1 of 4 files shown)


lib/nylas/handler/api_operations.rb
```diff
@@ -15,22 +15,22 @@ module Get
...
-        [response[:data], response[:request_id]]
+        [response[:data], response[:request_id], response[:headers]]
       end
 
       # Performs a GET call to the Nylas API for a list response.
       #
       # @param path [String] Destination path for the call.
       # @param query_params [Hash, {}] Query params to pass to the call.
-      # @return [Array(Array(Hash), String, String)] Nylas data array, API Request ID, and next cursor.
+      # @return [Array(Array(Hash), String, String, Hash)] Nylas data array, API Request ID, next cursor, and response headers.
```
AaronDDM, line 29: @kraju3 Looks like this is running into a lint error, can you move some of this text onto a new line? I think if you add spacing at the start of the newline, it should automatically concat it into the previous return statement.
```suggestion
      # @return [Array<Array<Hash>, String, String, Hash>]
      #   Nylas data array, API Request ID, next cursor, and response headers.response headers.
```

AaronDDM (approved):

PR #522, opened 2025-05-28
Implement encoding for path for special characters

kraju3:
# Description

# License


AaronDDM:
@kraju3 unfortunately this will not work all the time. Turns out, some `message ids` include `/` in them 🙃 - so I had to come up with a more robust solution for the nodejs SDK, we'll likely need to do that here as well:
https://github.com/nylas/nylas-nodejs/pull/642

PR #527, opened 2025-06-18
feat: add single_level query parameter support for Folders list endpoint

AaronDDM:
# Description
- feat: add single_level query parameter support for Folders list endpoint

# License


AaronDDM:
@kraju3 I think for ruby, for such cases, I am going to focus one improving the doc comments going forward - even though functionally nothing changes.

PR #533, opened 2025-09-17
Include headers in response

matiassalles99:
A few months ago we reached out to Kiran about the X-Retry-After header not being included in errors, so he made these changes https://github.com/nylas/nylas-ruby/pull/517

We are now trying to make use of these headers, but it looks like they're still not included, you can reproduce it with a similar script:

```ruby

error = nil
mutex = Mutex.new

threads = 50.times.map do
  Thread.new do
    res = Nylas::V3::GetThread.call(user:, thread_id:) # this is a service we have which just makes a get_thread call
    unless res.success?
      mutex.synchronize { error ||= res.error }
    end
  end
end

threads.each(&:join)

# then run
error.headers # -> will be nil
```

# Description
- The PR that Kiran opened assumed that the response param, which gets handed over from method to method, was the full response, but it's actually the response body
- Instead of refactoring most of the HttpClient class, I just added an optional headers param 

# License

### DIFF
--- a/lib/nylas/handler/http_client.rb
+++ b/lib/nylas/handler/http_client.rb
@@ -41,7 +41,7 @@ module Nylas
             content_type = response.headers["content-type"].downcase
           end
 
-          parsed_response = parse_json_evaluate_error(result.code.to_i, response.body, path, content_type)
+          parsed_response = parse_json_evaluate_error(result.code.to_i, response.body, path, content_type, response.headers)
           # Include headers in the response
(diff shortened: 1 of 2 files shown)


matiassalles99:
@AaronDDM @samuelpx is there anything I can help with to get this one merged?

AaronDDM:
Sorry @matiassalles99 - this totally slipped, on me. Taking a look. I will likely get it merged and a new release only next week however.

PR #546, opened 2026-07-21
[TA-3805] docs(calendars): document tentative_as_busy in get_free_busy request body

AaronDDM:
## Summary

Adds `tentative_as_busy` support to the `Calendars#get_free_busy` request body in the Ruby SDK, aligning it with the typed free/busy request models in the other Nylas SDKs (e.g. Python's `GetFreeBusyRequest`).

The Ruby SDK is entirely Hash-based and has no type system, so customers can already pass `tentative_as_busy` through the untyped `request_body` Hash. As with the other SDKs, this change makes the field a documented, first-class part of the request contract — the Ruby-idiomatic equivalent of a typed request model is documenting the accepted hash keys in the YARD comment (matching how `create`/`update` document their nested request fields).

## Changes

- Document the `get_free_busy` request body keys (`start_time`, `end_time`, `emails`, `tentative_as_busy`) in the YARD comment. `tentative_as_busy` uses wording consistent with the other SDKs: when set to `false`, treats tentative calendar events as `busy: false`; only applicable for Microsoft and EWS calendar providers; defaults to `true`.
- Fix the `identifier` param type in the same doc block (`str` → `String`).
- Add a spec asserting `tentative_as_busy` is forwarded in the request body.
- Add an `Unreleased` CHANGELOG entry.

## Testing

- `bundle exec rspec spec/nylas/resources/calendars_spec.rb` — 11 examples, 0 failures.
- `bundle exec rubocop lib/nylas/resources/calendars.rb spec/nylas/resources/calendars_spec.rb` — no offenses.

## Related

- TA-3805

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+### Unreleased
+* Documented `tentative_as_busy` support in the `Calendars#get_free_busy` request body, consistent with other Nylas SDKs
+
 ### [6.8.0]
 * Added Policies resource for managing application policies
 * Added Rules resource for managing inbox rules and listing rule evaluations
--- a/lib/nylas/resources/calendars.rb
+++ b/lib/nylas/resources/calendars.rb
@@ -90,8 +90,13 @@ module Nylas
 
     # Get the free/busy schedule for a list of email addresses.
     #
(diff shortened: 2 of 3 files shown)


qasim-nylas (approved):
LGTM 👍

